PR #8, opened 2019-08-29
Example conf.yaml always have OIDC KEYS set to readonly=False

peppelinux:


peppelinux:
Next week I'll add some screenshot into the readme and a better general organization.
Then I'll destroy all my environment, reclone  oidc-op and JWT-Connect-PythonRP, and run them following the instructions in Readme.md.

Once this be done the first point of the "plan" was accomplished :)

PR #9, opened 2019-08-29
Readme with some screenshots and someother instructions for a wider view

peppelinux:
Also embeds the howto found in the doc folder.
This would help guys to have a big view as soon as possible.


rohe:
The correct name is actually JWK Set (abbreviated JWKS).
Defined in https://tools.ietf.org/html/rfc7517#section-5

peppelinux:
Just ammended it, thx again

PR #16, opened 2019-11-20
added django-oidc-op registration api endpoint in conf

peppelinux:


peppelinux:
Examples conf updated for flask and django with

registration_api endpoint
introspection endpoint 

and PKCE support

PR #18, opened 2020-12-18
Develop

rohe:
Allow relative path in configuration to be made absolute.


peppelinux (approved):
Good to me,
for what's related to schema validation and data models, take a look here:
https://github.com/UniversitaDellaCalabria/uniCMS/blob/9f66c24aee16d6bdd7aee759e2ec5dcfca934cf9/src/cms/search/models.py#L7

I'm using pedantic, it works like a charm and it's very easy and eloquent in its messages.
Soon it became a must-have for me

PR #20, opened 2021-04-15
Include oidcendpoint in oidcop

rohe:
In the same manner as with oidcservice and OidcRP, oidcendpoint are in this release included in OidcOP.
Ease of maintenance is one reason. Keeping track of 3 packages is easier than keeping track on 4.
And OidcOP didn't contain a lot except for some examples.

While move oidcendpoint into OidcOP more things happened. One was that the new session management system
was included another that the new persistent storage model was applied.

So all and all quite a lot of changes.


peppelinux (approved):

rohe (commented):
kakor is the Swedish word for cookies, so yes ! :-)

PR #32, opened 2021-05-12
release-drafter & co also in master branch

peppelinux:
This should fix the current CI

feat: release-drafter also in master branch


rohe:
Do you understand why all tests fail. And they fail with the message that the directory tests doesn't exist.

peppelinux:
> Do you understand why all tests fail. And they fail with the message that the directory tests doesn't exist.

because in master we don't have tests folder.
I just removed the gh actions from this branch and removed the main branch from release-drafter configuration.

It just have to be merged in master, then release-drafter won't fail anymore ... My theory

rohe (approved):

Issue #22, opened 2021-04-27
disable weak algorithms


peppelinux:
The following algorithms must be disabled by default:

- RSA1_5
- none

peppelinux:
reference
https://tools.ietf.org/html/rfc8725

rohe:
OK

peppelinux:
in https://github.com/IdentityPython/oidc-op/pull/38 I found a good compromise

the weak algs doesn't been disabled by default but a warning message raises if there are found in the configuration

PR #38, opened 2021-05-15
construct_endpoint_info

peppelinux:
this PR helped understanding a weakness in the general configuration manager, used to put default capabilities and get it in provider-discovery json response BUT these paramenters doesn't really changed the endpoints behaviour.

With this approach I got several benefits, also did:

- chore: general refactor
- feat: WEAK algorithms makes a warning message in logs

this commit addresses https://github.com/IdentityPython/oidc-op/issues/22

### DIFF
--- a/src/oidcop/construct.py
+++ b/src/oidcop/construct.py
@@ -1,9 +1,15 @@
+import logging
+import re
+
 from functools import cmp_to_key
 
 from cryptojwt import jwe
 from cryptojwt.jws.jws import SIGNER_ALGS
 
 ALG_SORT_ORDER = {"RS": 0, "ES": 1, "HS": 2, "PS": 3, "no": 4}
+WEAK_ALGS = ['RSA1_5', 'none']
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


src/oidcop/construct.py
```diff
@@ -35,43 +41,45 @@ def assign_algorithms(typ):
...
+                    "Proposed set of values outside set of permitted, "
+                    f"'{attr}' sould be {_permitted} it's instead {_proposal}"
+                )
+            _info[attr] = _proposal
+        else:
+            if default_val is not None:
+                _info[attr] = default_val
+            elif "signing_alg_values_supported" in attr:
+                _info[attr] = assign_algorithms("signing_alg")
+                if "none" in _info[attr]:
```
peppelinux, line 71: @roland this is the only warning I'd tell you.
In you previour approach only if attr == "token_endpoint_auth_signing_alg_values_supported" the none value has been removed. I didn't understad this condition, so I remove none at every occourrence.

Feel free to revert and explain why that rule is only valid for `token_endpoint_auth_signing_alg_values_supported`

rohe (approved):

PR #39, opened 2021-05-15
jwks factory

peppelinux:
- fix: factory creates two different folder even If a private_path was defined in the general configuration
- chore: small code refactor with linting

With this PR now I have a clean base directory.
I configured custom folders where to store private and public files, but oidc-op still creates a private/token_jwks.json as a duplicate of which i configured with the custom private_path.

digging in the code I found what it could be a silly bug, now it's fixed

### DIFF
--- a/src/oidcop/token/handler.py
+++ b/src/oidcop/token/handler.py
@@ -139,27 +139,53 @@ def factory(server_get,
     TTYPE = {"code": "A", "token": "T", "refresh": "R"}
 
     key_defs = []
-    if code is not None:
-        key_defs.append({"type": "oct", "bytes": 24, "use": ["enc"], "kid": "code"})
-    if refresh is not None:
-        key_defs.append({"type": "oct", "bytes": 24, "use": ["enc"], "kid": "refresh"})
-    if token is not None:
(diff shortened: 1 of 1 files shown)


rohe (approved):

src/oidcop/token/handler.py
```diff
@@ -147,20 +147,42 @@ def factory(
...
+
+    for _keyd in key_defs:
+        if _keyd['kid'] == 'code':
+            code = _keyd
+        elif _keyd['kid'] == 'refresh':
+            refresh = _keyd
+        elif _keyd['kid'] == 'token':
+            token = _keyd
+
     if code is not None:
```
peppelinux, line 165: @rohe 
should it be `if not code` instead?

if these definitions already have been configured in the general conf, why should we have to overwrite them?
they would be created if those arguments are None and also configuration's key_defs are None.

Do you agree?
If it's the way Give me a sign of this, I'll push a PR for this

peppelinux: Ok, you've already achieved this here
https://github.com/IdentityPython/oidc-op/commit/062dc1dff66d215b698bff0bb8697f8d653e1d57

well done.

PR #41, opened 2021-05-15
Test 00 fixed and small cleanups

peppelinux:
fix: test 00 
fix: scope2claims -> `map` is a primitive and must not be overloaded as a variable

### DIFF
--- a/tests/op_config.json
+++ b/tests/op_config.json
@@ -222,7 +222,7 @@
   "id_token": {
     "class": "oidcop.id_token.IDToken",
     "kwargs": {
-      "default_claims": {
+      "base_claims": {
         "email": {
           "essential": true
         },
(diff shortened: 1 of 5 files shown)


tests/op_config.json
```diff
@@ -222,7 +222,7 @@
   "id_token": {
     "class": "oidcop.id_token.IDToken",
     "kwargs": {
-      "default_claims": {
```
peppelinux, line 225: I still see a disalignment on which variable name we must use.
I found this in op_config and base_claims in test.

I'm using base_claims, no problem to switch to default_claims, just need to decide which one to use and then a final alignment of code and documentation must be done

rohe: Should always be base_claims. I think I went through the code base yesterday and fixed that everywhere.

rohe (approved):

Issue #36, opened 2021-05-14
Session dump inconsistent state on some conditions


peppelinux:
@develop branch

If a first session have been done, the next one fails at verify_user view.
the failure is cause by the fact that the flushed session database doesn't have a 'db' element in it

the patchy workaroud is this, in verify_user view:

````
    session_manager = ec.endpoint_context.session_manager
    dump = session_manager.dump()
    if not dump.get('db'):
        dump['db'] = {}
        session_manager.load(dump)
````

GHere the traceback of the error

````
Traceback (most recent call last):
  File "/path/to/DEV/IdentityPython/OIDC/env/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/path/to/DEV/IdentityPython/OIDC/env/lib/python3.8/site-packages/django/core/handlers/base.py", line 179, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/path/to/DEV/IdentityPython/OIDC/env/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/path/to/DEV/IdentityPython/OIDC/django-oidc-op/example/oidc_provider/views.py", line 303, in verify_user
    _session_id = session_manager.create_session(
  File "/path/to/DEV/IdentityPython/OIDC/env/lib/python3.8/site-packages/oidcop/session/manager.py", line 189, in create_session
    _usi = self.get([user_id])
  File "/path/to/DEV/IdentityPython/OIDC/env/lib/python3.8/site-packages/oidcop/session/database.py", line 99, in get
    raise InconsistentDatabase('Missing session db')
oidcop.session.database.InconsistentDatabase: Missing session db
````

here the commented workaround in the example project
https://github.com/peppelinux/django-oidc-op/commit/e97963c02afd4dee3eddcd8f8bee06e0b0dec3ff

peppelinux:
this problem is still present

rohe:
Where in django-oidc-op do you call the dump, load, flush methods the oidc-op provides ?

peppelinux:
load in the endpoint views, example here BUT they're going to be refactored as ClassViews with method decorators
https://github.com/peppelinux/django-oidc-op/blob/426003dc8c963b944b5e95431094335d8f9e9465/oidc_provider/views.py#L393

[...]

PR #46, opened 2021-05-16
django_op example updated with develop branch

peppelinux:
django op example updated to oidc-op develop branch


rohe (approved):
If you're OK with this I trust you.

peppelinux:
We could decide to do this way:

Once we have v2.0.0 I'll remove django_op example code and put there only a README that points to django-oidc-op github repository. django-oidc-op could be migrated to idpy for a better reference, as well.

django-oidc-op releases Will match the major release value of oidc-op, so once we'll have the stable release of oidc-op then I'll publish django-oidc-op v2.x.y

@c00kiemon5ter does this sound good to you?

@rohe hope to meet your taste too. The only awarness I have is that, actually, django-oidc-op is the only implementation of a persistent storage based on dump()/load()/flush() methods of oidc-op session_manager, and still have this issue:
https://github.com/IdentityPython/oidc-op/issues/36

it doesn't mean that we have a real bug in oidc-op but just that we still need to exemplify a good (official) method to implement this feature in a clear, documentable, way

I can go further with a general refactor of django-oidc-op (based on ClassView) then try to visualize more closer that problem

rohe:
I'm fine with what you propose.
Hard to nail down #36 since I can't replicate it in my setup.

peppelinux:
with https://github.com/IdentityPython/oidc-op/pull/59 is now all ok to me,
once we get merged https://github.com/IdentityPython/oidc-op/pull/59 in develop branch I'll move django_op example as a reference to my repository (with a view to migrating it to idpy for a more robust reference).

So I close this for now, let's merge dpop2 and sman_flush and then I'll do this chore (otherwise I get a quite mess with too much branches!)

PR #51, opened 2021-05-16
Long road to coverage

peppelinux:
chore: some more points up to a better coverage

some pragma no-cover have been used, and some "TODO" notes as a reminder, some tests have been created

### DIFF
--- a/src/oidcop/utils.py
+++ b/src/oidcop/utils.py
@@ -7,7 +7,7 @@ import sys
 import yaml
 
 
-def load_json(file_name):
+def load_json(file_name): # pragma: no cover
     with open(file_name) as fp:
         js = json.load(fp)
     return js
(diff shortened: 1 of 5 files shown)


src/oidcop/utils.py
```diff
@@ -19,7 +19,7 @@ def load_yaml_config(file_name):
     return c
 
 
-def yaml_to_py_stream(file_name):
+def yaml_to_py_stream(file_name): # pragma: no cover
```
peppelinux, line 22: all these utilities are not used in the project, so I'm ok to keep them for future needs but marked as uncovered to give them a minor relevance in the code

rohe: I'm actually OK with weeding them out.

rohe (approved):

PR #54, opened 2021-05-18
Coverage improved

peppelinux:
* chore: coverage of user_authn.user increased
* chore: src/oidcop/token/id_token.py coverage from 86% to 90%

@rohe I made a code cdoverage in a fast and dirty way, this code is questionable, your quick review or commits above this may make object access methods more obvious, which I modestly dug up with the debugger

Actually `src/oidcop/user_info/__init__.py` is the file that keep the coverage low, covering this will get a better result ... code is similar to our society, improving the conditions of those who are worse off ... :)

### DIFF
--- a/tests/test_05_jwt_token.py
+++ b/tests/test_05_jwt_token.py
@@ -86,6 +86,7 @@ MAP = {
     "authorization_code": "code",
     "access_token": "access_token",
     "refresh_token": "refresh_token",
+    "id_token": "id_token"
 }
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-django_op/db.sqlite3
-templates
 static/
 private/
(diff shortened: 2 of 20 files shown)


peppelinux:
Ah, I also added the templates folders that was missing, the flask example didn't worked without these

tests/test_05_jwt_token.py
```diff
@@ -86,6 +86,7 @@
     "authorization_code": "code",
     "access_token": "access_token",
     "refresh_token": "refresh_token",
+    "id_token": "id_token"
```
peppelinux, line 89: @rohe probably I forget this there ... It's not needed

rohe:
I think we should move userpassjinja2 out of the core code and into the example tree.
But we can do that later.

PR #56, opened 2021-05-20
Dpop2

rohe:
Worked through the DPoP support.

OAuth 2.0 Demonstrating Proof-of-Possession at the Application Layer (DPoP)
draft-ietf-oauth-dpop-03


peppelinux:
sounds great Roland, can you fix the conflict in src/oidcop/oauth2/device_authorization.py?

We should improve this before merge: https://github.com/IdentityPython/oidc-op/pull/54
this will increase about 2% but please take a look at the code, it can't be merged as it is, do your revision and tell me how to make change or merge and then do your commits, you drive.

overall, this PR looks awesome, ready for merge but: conflicts first.
I can do that but I think that you're already in the flow with this

peppelinux (approved):

src/oidcop/session/grant.py
```diff
@@ -27,11 +28,11 @@ class GrantMessage(ImpExp):
...
 
     def __init__(
-        self,
-        scope: Optional[str] = "",
-        authorization_details: Optional[dict] = None,
-        claims: Optional[list] = None,
-        resources: Optional[list] = None,
+            self,
+            scope: Optional[str] = "",
+            authorization_details: Optional[dict] = None,
```
peppelinux, line 33: This means that oidc-op is also aware of 
https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rar

@rohe this should be exalted in the README/Official Docs
do you agree to add this feature/compliance to Documentation introduction, where we had all the specs compliances of oidc-op?

rohe: Yes!

peppelinux: here https://github.com/IdentityPython/oidc-op/pull/58

peppelinux:
I think that's ready to be merged, @nsklikas do you agree?
go @rohe, go!

peppelinux:
@rohe maybe some unit tests can be done if you agree :)

peppelinux (approved):

src/oidcop/oauth2/add_on/dpop.py
```diff
@@ -112,36 +114,35 @@ def post_parse_request(request, client_id, endpoint_context, **kwargs):
...
-    except KeyError:
-        endpoint_context.cdb[client_id]["dpop_jkt"] = {_jkt: _dpop.key}
-
     # Need something I can add as a reference when minting tokens
-    request["dpop_jkt"] = _jkt
+    request["dpop_jkt"] = as_unicode(_dpop.key.thumbprint("SHA-256"))
     return request
 
 
 def token_args(endpoint_context, client_id, token_args: Optional[dict] = None):
```
peppelinux, line 122: token_args is at the same time the name of the function and the name of an argument.
We should avoid naming collision, do you agree?

peppelinux (changes requested):

rohe:
Included in #65

src/oidcop/oauth2/add_on/dpop.py
rohe: Yep!

Issue #53, opened 2021-05-18
[Session manager] Missing key and salt after flush and load


peppelinux:
Once I flushed the session manager, sometimes, on the next request I miss key and salt.

````
INFO:oidc_provider.views:Request at the "userinfo" endpoint
[2021-05-18 08:20:15,588] Response params: {
  "email": "sdf@sdf.it",
  "sub": "a7b0dea2958aec275a789d7d7dc8e7d09c6316dd4fc6ae92742ed3297e14dded"
}
 [(DEBUG) oidc_provider.views.do_response:68]
DEBUG:oidc_provider.views:Response params: {
  "email": "sdf@sdf.it",
  "sub": "a7b0dea2958aec275a789d7d7dc8e7d09c6316dd4fc6ae92742ed3297e14dded"
}

[2021-05-18 08:20:15,590] Missing salt or key in session dump [(CRITICAL) oidc_provider.models.load:445]
CRITICAL:oidc_provider.models:Missing salt or key in session dump

````

![image](https://user-images.githubusercontent.com/1297620/118617521-04162d00-b7c3-11eb-8a11-d6155005d41f.png)

It needs to be investigated more closely

peppelinux:
fixed here: https://github.com/IdentityPython/oidc-op/pull/65

Issue #50, opened 2021-05-16
[Documentation] Pairwise salt


peppelinux:
the salt involved in subject must be documented, otherwise it's hard to go in production with multiple workers that have random salt in each of their instances

see `oidcop.session.manager.PairWiseID` and other IDs

rohe:
Salt is defined as a dump parameter in SessionManager. It should take care of this.
May have to add a test to verify it.

peppelinux:
it seems to me that's not actually aware of it

````
      },
      "session_key": {
        "filename": f"{OIDC_JWKS_PRIVATE_PATH}/session_jwk.json",
        "type": "OCT",
        "use": "sig"
      },
      "sub_func": {
        "public": {
          "class": "oidcop.session.manager.PublicID",
          "kwargs": {
            "salt": "sdfsdfdsf"
          }
        },
        "pairwise": {
          "class": "oidcop.session.manager.PairWiseID",
          "kwargs": {
            "salt": "sdfsdfsdf"
          }
        }
     },
 ````
 
 the dump shows random salt instead of the defined one

peppelinux:
and more, it happens, somethimes, that session manager dump doesn't have salt and key correctly valued

````
...

[...]

Issue #47, opened 2021-05-16
Missing TokenIntrospection in examples project


peppelinux:
It would be better to have it :)

rohe:
example/flask_op do have an introspection endpoint.
Don't know which example project you are talking about.

peppelinux:
The same but I miss It, my excuses 😎

peppelinux:
done here:
https://github.com/IdentityPython/oidc-op/pull/59/commits/4286cc8a9d76c4da094ea756a9c9ecb1897ca84e

PR #59, opened 2021-05-21
fix: Session dump inconsistent schema after sman flush

peppelinux:
- fixes https://github.com/IdentityPython/oidc-op/issues/36
- fixes https://github.com/IdentityPython/oidc-op/issues/53
- fixes https://github.com/IdentityPython/oidc-op/issues/50
- fixes https://github.com/IdentityPython/oidc-op/issues/47
- chore: small refactor of sman init

and also
- fix: session manager gets password from general config
- fix: session manager salt optionally from global conf
- chore: session Database init small generalization


rohe:
Realised this should be done a couple of days ago. Glad you did it for me :-)

rohe:
But you should fix the tests.

peppelinux:
@rohe I fixed the tests ... rather I fixed the pointers in the session database.
I also tested static password and salt in globalconf, it works good as if they were randomic.
Something more, something better than before, thank to bugs!

src/oidcop/session/manager.py
```diff
@@ -105,6 +101,21 @@ def __init__(
...
+        return self.key
+
+    def load_salt(self):
+        self.salt = self.conf.get("salt") or rndstr(32)
+        return self.salt
+
+    def _init_db(self):
+        Database.__init__(
+                    self,
+                    key=self.key or self.load_key(),
```
peppelinux, line 115: This make it safer when one or multiple flush occours.
The schema and the init values will be always consistent, and more, the flush doesn't overwrite the class attributes values or any other thing that then MUST be rebuilted as it were before

src/oidcop/session/manager.py
```diff
@@ -470,7 +483,11 @@ def remove_session(self, session_id: str):
     def local_load_adjustments(self, **kwargs):
         self.crypt = Crypt(self.key)
 
+    def flush(self):
+        super().flush()
+        self._init_db()
+
 
-def create_session_manager(server_get, token_handler_args, sub_func=None):
+def create_session_manager(server_get, token_handler_args, sub_func=None, conf=None):
```
peppelinux, line 491: let the configuration be received by its admirers!

peppelinux:
It also fixed https://github.com/IdentityPython/oidc-op/issues/53

rohe:
We (I) probably should document the three methods dump/flush/load better.
Here is a start:

The bases for these three methods are the `parameter` definitions.
Those are the attributes that are affected by the three methods.

`dump` returns the value in a format that is possible to json dump
`flush` assigns as value to the attribute the value that is defined in the parameter specification.
`load` assigns a value to the attribute.

This means that after a `flush` you MUST do a `load` to get a working setup.

Having something like salt being randomly defined means that all entities that needs to have the same salt must be loaded with the same state. This means that having anything that should be known to more than one worker being assigned randomly is probably a bad idea. You can make it work by initialising one worker, dump its initial state and then have all the subsequent initialised workers being loaded with the first ones state.

src/oidcop/session/manager.py
rohe: But the dump/(flush)/load is not there to have something being built as it where before.
It's there to assign a specific state.
If that state is some initial state that you want to rebuilt so be it, but that is just one possible usage.

peppelinux:
We (I) should improve this commit making load_key and load_salt aware of the random value, stored elsewhere in two private attributes of sman objects, this Will prevent a random value Will be generated After any flush.

But multiple workers Will Always have their own randomic, that's why I put the static password and salt parameters in the global configuration.

These will prevent any randomness.

So, what to do know with this PR?

peppelinux:
salt value in parameters dict Is "".
This caused that salt being "" After a flush without any load (It May happen in some condition) and self.salt, the originari one, were overwritten by this "" runtime

The static password/Key configurations fixed this behaviour by-default and an additional commit can make a immutabile random salt and Key value instead of calling eberytime the rnd() finction

src/oidcop/session/manager.py
peppelinux: I agree to have randomness only on start and then reuse Always that One. This Is a trivial commit.

I really don't want to use a dump from the First worker but instead those parameters in the global configuration

rohe:
You should NEVER do a flush without a load. That an invitation to disaster !
`flush` resets all defined parameters to their default value (defined in parameter).
Default values all are empty values: "", [], {}, 0, False, ..

rohe:
If you don't want random values per worker, then the value MUST be defined in the configuration (basically being static) or in a file referenced by the configuration (a la session_key).

peppelinux:
> You should NEVER do a flush without a load. That an invitation to disaster !
> `flush` resets all defined parameters to their default value (defined in parameter).
> Default values all are empty values: "", [], {}, 0, False, ..

It could happen due to unhandled exceptions.
With this PR I can make many flush without breaking anything.

This PR also enable static password and salt.

I'llonly do a another commit to improve the init_db method as we shared in the previous posts.

Can I intend that we agree and then go ahead?

peppelinux:
Once this PR Will be merged then se could see this
https://github.com/IdentityPython/oidc-op/issues/60

session_key, sub_funcs, password and salts MAY be under a session manager umbrella, as a opconfiguration root mode and not sparse as they are. Hope to meet your agreement, this Will make Easier to read/understand the configuration and obviously make a Better documentation 💪

rohe:
I think we must be clear about what is under the dump/flush/load umbrella and what isn't.
If we allow special handling of some attributes that implements something in-between persistent/non-persistent storage we are not making it any easier either for us or for users of the package.

peppelinux:
They all are related to session manager, Just this. All this parameters belongs to session manager.

But this aspect Is out of scope in this PR, i talked too much on things not related to the scope of this PR. It Just fix few small bugs and now my implementation of storage, in django-oidc-op develop branch, works great.

This afternoon I'll add another commit to follow what shared here between US but feel free to suggest, tell, do what you intende to be Better for the roadmap. I accept task 😊

peppelinux:
Ok, made it, unit tests works great and also added another test to deal with the issues it fixes and the feature I want to push with this contribution:

- https://github.com/IdentityPython/oidc-op/issues/36
- https://github.com/IdentityPython/oidc-op/issues/53
- key and salt, even if randomic, can be flushed BUT cannot be overwritten runtime. Even if they are randomic they will be immutable runtime. So I added two private attributes, _key and _salt and a RO check as __setattr__

image

src/oidcop/session/manager.py
peppelinux: and more ... It's something that belongs to security. A project password/salt should not be changed by code runtime. If this change is needed it would be possibile to change this parameter in configuration (or leave it random as it would is) and do a gracefull restart of the service.

my two cents relating the feature of getting _key and _salt some private, readonly values.
We can revert, sure, but I put it there as a security feature

src/oidcop/session/manager.py
peppelinux: ok, I intend that you agree at least this :)
closing.

peppelinux:
Last commit improves Documentation and definitively closes https://github.com/IdentityPython/oidc-op/issues/50

rohe (approved):

peppelinux:
I think that conf dict should be readonly.
@rohe do we have to do something in Server configuration and OPconfiguration class?

Once it's created a lock state should prevent any other changes in the future

rohe:
Agree that it should be read only.
Realised that we didn't have support for token endpoint in an OAuth2 Authorization Server so working on it right now. Meant among other things that I had to create ASConfiguration.

Not sure how you would enforce a lock. If someone wanted to get around it, it would be hard to stop.
But it had to demand some work, couldn't be done by default which I guess is as far as we can get.

rohe:
Included in #65

PR #64, opened 2021-05-23
Added token endpoint to oidc-op/oauth2.

rohe:
As a follow up on the above:

- Added OAuth2ClaimsInterface
- Added ASConfiguration
- Changed name of tests to make clear if they where oauth2 or oidc based.
- Made all OAuth2/OIDC based tests use ASConfiguration/OPConfiguration.

### DIFF
--- a/src/oidcop/client_authn.py
+++ b/src/oidcop/client_authn.py
@@ -131,9 +131,7 @@ class ClientSecretPost(ClientSecretBasic):
 
     def verify(self, request, **kwargs):
         if (
-            self.server_get("endpoint_context").cdb[request["client_id"]][
-                "client_secret"
-            ]
+            self.server_get("endpoint_context").cdb[request["client_id"]]["client_secret"]
             == request["client_secret"]
         ):
             return {"client_id": request["client_id"]}
@@ -148,9 +146,7 @@ class BearerHeader(ClientSecretBasic):
     tag = "bearer_header"
 
     def is_usable(self, request=None, authorization_token=None):
-        if authorization_token is not None and authorization_token.startswith(
-            "Bearer "
-        ):
+        if authorization_token is not None and authorization_token.startswith("Bearer "):
             return True
         return False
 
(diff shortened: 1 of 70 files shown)


peppelinux:
it was 2018-2019 when I had talked you to divide OAuth2 profile from OIDC one. You did it finally, great!

src/oidcop/client_authn.py
```diff
@@ -203,9 +199,7 @@ def verify(self, request, key_type, **kwargs):
         if _sign_alg and _sign_alg.startswith("HS"):
             if key_type == "private_key":
                 raise AttributeError("Wrong key type")
-            keys = _context.keyjar.get(
-                "sig", "oct", ca_jwt["iss"], ca_jwt.jws_header.get("kid")
-            )
+            keys = _context.keyjar.get("sig", "oct", ca_jwt["iss"], ca_jwt.jws_header.get("kid"))
```
peppelinux, line 202: I'm wondering why you brought all the pep8 modifications back to their original state but if you like it more so for me it's ok and totally relative.

Right it occurs to me that this revision will be more challenging;)

src/oidcop/configure.py
```diff
@@ -241,8 +227,73 @@ def __init__(
...
+        self.claims_interface = None
+        self.cookie_handler = None
+        self.endpoint = {}
+        self.httpc_params = {}
+        self.issuer = ""
+        self.keys = None
+        self.session_key = None
+        self.template_dir = None
+        self.token_handler_args = {}
+        self.userinfo = None
```
peppelinux, line 287: and these are OP's

        self.add_on = None
        self.authz = None
        self.authentication = None
        self.base_url = ""
        self.capabilities = None
        self.cookie_handler = None
        self.endpoint = {}
        self.httpc_params = {}
        self.id_token = None
        self.issuer = ""
        self.keys = None
        self.login_hint2acrs = {}
        self.login_hint_lookup = None
        self.session_key = None
        self.sub_func = {}
        self.template_dir = None
        self.token_handler_args = {}
        self.userinfo = None
        self.password = None
        self.salt = None

They are replicated from OPconfiguration, in a way that if we should add a argument we then need to put in OPConf and at the same time in ASConf. I think that it would be better to have a BaseClass to inherit and take all these attribute from it instead of repeat them

peppelinux (changes requested):
In a single commit we have changes of a different nature. I would have definitely split the code formatting style commit versus adding the additional functionality. It is difficult to review changes to 74 files in a single commit.

Anyway the only change I feel like asking you is to avoid replicating the OPConfiguration code in ASConfiguration as a copy / paste. The parts they have in common must be inherited.

Starting from OAuth2 as a basis we could do to inherit ASConfiguration in OPConfiguration and cut the bull's-head!

src/oidcop/configure.py
rohe: Done

src/oidcop/client_authn.py
rohe: Yeah, I shouldn't have run black on it.

peppelinux: Not without another commit in between 😜

rohe: Yeah!

PR #65, opened 2021-05-23
Consolidated Pull Request for v2.0.0 development

peppelinux:
This PR merges

- token endpoint to oidc/oauth2 (https://github.com/IdentityPython/oidc-op/pull/64)
- sman_flush (https://github.com/IdentityPython/oidc-op/pull/59) that itself carries a lot of sparse things
- Dpop2 (https://github.com/IdentityPython/oidc-op/pull/56)

it fixed merge conflicts, and tested all the things that actually seems to works fine.
I did this job to get an idea of ​​how to continue development in view of the v2.0.0 release, without scattering too much branches.

Since it works and I've done the job, I decided to share it anyway.

@nsklikas currently your opinion is important, if you can not do the revision please unlock us with a sign from you, if we continue to disperse on different branches we will then lose further time to combine all the pieces.

Concluding the remaining issues on the 2.0.0 milestone and following a few days of testing and rereading the features and documentation, we would definitively close this phase.

and, not at last, Documentatioooon!!!!


rohe (approved):

peppelinux:
Mhhh... the last commit in token branch makes problem in unit tests @rohe

example/flask_op/config.json
```diff
@@ -185,8 +185,11 @@
           "class": "oidcop.oauth2.introspection.Introspection",
           "kwargs": {
             "client_authn_method": [
-              "client_secret_post"
-            ],
+              "client_secret_post",
+              "client_secret_basic",
+              "client_secret_jwt",
+              "private_key_jwt"
+            ]
```
nsklikas, line 192: Missing comma: `],`

src/oidcop/oauth2/add_on/dpop.py
```diff
@@ -58,6 +59,7 @@ def create_header(self) -> str:
         payload = {k: self[k] for k in self.body_params}
         _jws = JWS(payload, alg=self["alg"])
         _headers = {k: self[k] for k in self.header_params}
+        self.key.kid = ""
```
nsklikas, line 62: Why is this needed?

example/flask_op/config.json
peppelinux: please do a pull, I fixed it just now

src/oidcop/oauth2/add_on/dpop.py
peppelinux: actually it should be rfc7836 aware
https://github.com/IdentityPython/JWTConnect-Python-CryptoJWT/blob/815d5b6a3d80852fcaaff5f8227f735124e06f2c/src/cryptojwt/jwk/__init__.py#L274

src/oidcop/oidc/token.py
```diff
@@ -109,20 +35,25 @@ def process_request(self, req: Union[Message, dict], **kwargs):
...
-            )
+            return self.error_cls(error="invalid_request", error_description="Missing code")
 
         _session_info = _mngr.get_session_info_by_token(_access_code, grant=True)
         grant = _session_info["grant"]
 
+        token_type = "Bearer"
+
+        # Is DPOP supported
+        if "dpop_signing_alg_values_supported" in _context.provider_info:
```
nsklikas, line 51: I think that it would be clearer if we had a flag for enabling DPoP instead. E.g. `dpop_enabled`.

Of course that flag shouldn't be exposed in the provider_info.

src/oidcop/session/claims.py
```diff
@@ -119,7 +113,7 @@ def get_claims(self, session_id: str, scopes: str, usage: str) -> dict:
 
     def get_claims_all_usage(self, session_id: str, scopes: str) -> dict:
         _claims = {}
-        for usage in ["userinfo", "introspection", "id_token", "access_token"]:
+        for usage in self.claims_types:
             _claims[usage] = self.get_claims(session_id, scopes, usage)
```
nsklikas, line 117: This should be wrapped in a try except:
```python
try:
    _claims[usage] = self.get_claims(session_id, scopes, usage)
except ServiceError:
    pass
```
OR `_get_module` must not raise and return `None`

src/oidcop/oidc/token.py
rohe: OK, we can put in endpoint_context.

src/oidcop/oauth2/add_on/dpop.py
rohe: Because you don't want it go looking for a key with a specific `kid`.
This since the key you get in the headers doesn't have a `kid`.

src/oidcop/session/claims.py
rohe: None has a special meaning. Which is that the `usage` doesn't come with any claims release.
An example of such a `usage` is authorization_code.

ServiceError on the other hand means: you are asking for claims connected to a specific `usage` I should know about but don't or can't handle.

src/oidcop/oidc/token.py
rohe: Pushed a modification with this.

nsklikas (approved):

PR #66, opened 2021-05-24
chore: pick_auth refactor

peppelinux:
Small refactor for a better readability

### DIFF
--- a/src/oidcop/oauth2/authorization.py
+++ b/src/oidcop/oauth2/authorization.py
@@ -429,11 +429,16 @@ class Authorization(Endpoint):
         if auth_id:
             return _context.authn_broker[auth_id]
 
+        res = None
         if acr:
             res = _context.authn_broker.pick(acr)
         else:
-            res = pick_auth(_context, request)
-
+            try:
+                res = pick_auth(_context, request)
+            except Exception as e:
+                logger.debug(
+                    f"An error occurred while picking the authN broker: {exc}"
+                )
         if res:
             return res
         else:
--- a/src/oidcop/oauth2/authorization.py
+++ b/src/oidcop/oauth2/authorization.py
@@ -435,8 +435,8 @@ class Authorization(Endpoint):
         else:
             try:
                 res = pick_auth(_context, request)
-            except Exception as e:
-                logger.debug(
+            except Exception as exc:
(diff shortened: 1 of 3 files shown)


src/oidcop/user_authn/authn_context.py
```diff
@@ -108,58 +108,45 @@ def default(self):
             return None
 
 
-def pick_auth(endpoint_context, areq, all=False):
+def pick_auth(endpoint_context, areq, pick_all=False):
```
peppelinux, line 111: all is a python primitive, better to avoid to overwrite it even if in this frame only

rohe (approved):

src/oidcop/oauth2/authorization.py
```diff
@@ -429,11 +429,16 @@ def pick_authn_method(self, request, redirect_uri, acr=None, **kwargs):
...
         if acr:
             res = _context.authn_broker.pick(acr)
         else:
-            res = pick_auth(_context, request)
-
+            try:
+                res = pick_auth(_context, request)
+            except Exception as e:
+                logger.debug(
+                    f"An error occurred while picking the authN broker: {exc}"
```
nsklikas, line 440: Can an exception be raised here?

peppelinux: Really don't know, I Just adapt the old code and take It as It was but moving in different context

@rohe do we have tests for this? I think so

peppelinux: Ah, typo error, e and not ecco!

peppelinux: paradoxically we are sure that the current tests do not fail, otherwise the typo error would have created a higher order exception

peppelinux: fixed: https://github.com/IdentityPython/oidc-op/pull/66/commits/c69fbfcd0a93a95d79ba5bd9a0327731c8a920dd

peppelinux: @nsklikas does it sounds good to you or should we still improve this?

rohe: Seems to be covered by the tests.

Issue #33, opened 2021-05-13
Session manager dump, missing values


peppelinux:
Hi there,
The following is a dump of a session, the elements in bold should catch our attention

{
  "expires_at": 1620989852,
  "issued_at": 1620946652,
  "not_before": 0,
  "revoked": false,
  "usage_rules": {
    "authorization_code": {
      "supports_minting": [
        "access_token",
        "refresh_token",
        "id_token"
      ],
      "max_usage": 1
    },
    "access_token": {},
    "refresh_token": {
      "supports_minting": [
        "access_token",
        "refresh_token"
      ]
    }
  },
  "used": 2,
  "authentication_event": {
    "oidcop.authn_event.AuthnEvent": {
      "uid": "wert",
      "authn_info": "oidcop.user_authn.authn_context.INTERNETPROTOCOLPASSWORD",
      "authn_time": 1620946652,
      "valid_until": 1620950252
    }
  },
  "authorization_request": {
    "oidcmsg.oidc.AuthorizationRequest": {
      "redirect_uri": "https://{domain}:{port}/authz_cb/django",
      "scope": "openid profile email address phone",
      "response_type": "code",
      "nonce": "f5rAJZdZ5zOpMv1wSzL7qFOO",
      "state": "pEoiWGl4BG7WrvYyK34JhV6BoMJtlnQ9",
      "code_challenge": "PsfTtRje76sGk4muOhnXD1rjI10fNKwYNLgsLcawgrw",
      "code_challenge_method": "S256",
      "client_id": "VjIX06LEI6MCrlRL-m6Hvg"
    }
  },
  "claims": {
    "userinfo": {
      "sub": null,
      "name": null,
      "given_name": null,
      "family_name": null,
      "middle_name": null,
      "nickname": null,
      "profile": null,
      "picture": null,
      "website": null,
      "gender": null,
      "birthdate": null,
      "zoneinfo": null,
      "locale": null,
      "updated_at": null,
      "preferred_username": null
    },
    "introspection": {},
    "id_token": {
      "email": null,
      "email_verified": null
    },
    "access_token": {}
  },
  "resources": [
    "VjIX06LEI6MCrlRL-m6Hvg"
  ],
  "scope": [
    "openid",
    "profile",
    "email",
    "address",
    "phone"
  ],
  "sub": "1a4962cd184469545b6149f35485b407ccb21a8a311ed78f5374171cb1d66b2c",
  "issued_token": [
    {
[...]

PR #76, opened 2021-05-30
Find scope

rohe:
This is about finding the scope for a token. The scope may be in the token, in a token in a hierarki of tokens used to mint the token in question or in the grant that is minting/have minted the token.

The idea is that you follow the path from the token through the set of based_on tokens to end up at the grant.
The first item that has a scope defined decides the outcome.


peppelinux (approved):

peppelinux:
Does this PR close this issue
#33
?

I havent analyzed yet the dump

rohe:
Yes, exactly!

PR #77, opened 2021-05-31
Make id token lifetime configurable

nsklikas:
Seems like this was broken when id token handling was moved to the token handler


peppelinux (approved):
Completely agree

rohe (approved):

PR #81, opened 2021-06-01
Fix no session

### LINKED ISSUE #78: ID Token introspection
Currently providing an ID Token to the introspection endpoint results in an unhandled exception.

@rohe @peppelinux What are out plans regarding using ID tokens in the introspection endpoint? 
IMO we should allow introspection of ID tokens.

nsklikas:
Closes #78


peppelinux (approved):
completely agree :ok_hand:

rohe (approved):

PR #82, opened 2021-06-01
Fix bug from token semantics refactor

nsklikas:
Looks like we missed this bug.

Also the tests didn't catch this.


peppelinux (approved):
Good catch!

rohe (approved):

PR #94, opened 2021-06-09
Handle all exceptions

nsklikas:
I encounter a bug were the exception defined in https://github.com/IdentityPython/oidc-op/blob/master/src/oidcop/token/__init__.py#L117 was raised when I was trying to exchange an authorization code for an access token. 

Perhaps we should handle all exceptions there? Or simply create an OidcopError exception from which all other exceptions will inherit and we can simply catch that?

What do you think?

Closes https://github.com/IdentityPython/oidc-op/issues/97

### DIFF
--- a/src/oidcop/token/handler.py
+++ b/src/oidcop/token/handler.py
@@ -1,7 +1,7 @@
 import logging
 import os
-from typing import Optional
 import warnings
+from typing import Optional
 
 from cryptojwt.exception import Invalid
 from cryptojwt.key_jar import init_key_jar
@@ -13,6 +13,7 @@ from oidcop.exception import InvalidToken
 from oidcop.token import DefaultToken
 from oidcop.token import Token
 from oidcop.token import UnknownToken
+from oidcop.token import WrongTokenClass
 from oidcop.token import WrongTokenType
 from oidcop.util import importer
 
@@ -73,7 +74,15 @@ class TokenHandler(ImpExp):
         for typ in order:
             try:
                 res = self.handler[typ].info(token)
-            except (KeyError, WrongTokenType, InvalidToken, UnknownToken, Invalid, AttributeError):
+            except (
+                KeyError,
+                WrongTokenType,
+                InvalidToken,
+                UnknownToken,
+                Invalid,
+                AttributeError,
+                WrongTokenClass,
+            ):
                 pass
             else:
                 return self.handler[typ], res


rohe:
Having said that we still will have exception that belong to OidcMsg and CryptoJWT cropping up in OidcOP.

peppelinux:
Can you share the test that make this exception?
I'm interested about why this exception was not handled by unit tests

nsklikas:
Ok I looked some more into it and it seems that the bug is elsewhere.

I try to run a regular code flow. When I try to exchange the authorization code with an access_token the following error is raised:
```
  File "/srv/venv/lib/python3.7/site-packages/oidcop/oauth2/token.py", line 407, in process_request
    _access_token, grant=True
  File "/srv/venv/lib/python3.7/site-packages/oidcop/session/manager.py", line 459, in get_session_info_by_token
    _token_info = self.token_handler.info(token_value)
  File "/srv/venv/lib/python3.7/site-packages/oidcop/token/handler.py", line 55, in info
    _handler, item_info = self.get_handler(item, order)
  File "/srv/venv/lib/python3.7/site-packages/oidcop/token/handler.py", line 75, in get_handler
    res = self.handler[typ].info(token)
  File "/srv/venv/lib/python3.7/site-packages/oidcop/token/__init__.py", line 117, in info
    raise WrongTokenClass(_res["token_class"])
oidcop.token.exception.WrongTokenClass: access_token
```

The bug is cause from the token minting. In https://github.com/IdentityPython/oidc-op/blob/0f92cfbf8079224c4c11e861a6c93d0efcf96d61/src/oidcop/session/grant.py#L306 it calls `Token` without passing the `token_class`. The function called is https://github.com/IdentityPython/oidc-op/blob/0f92cfbf8079224c4c11e861a6c93d0efcf96d61/src/oidcop/token/__init__.py#L74
Which expects a token_class and if it isn't present then it defaults to `authorization_code`. It's weird that we haven't stumble on it all this time.

nsklikas:
I will close this PR and open an issue so that we can further investigate it

nsklikas:
> Having said that we still will have exception that belong to OidcMsg and CryptoJWT that will crop up in OidcOP.

To solve this we can have a common exception base that will be shared across the libraries, e.g.:
```
OIDCError
OidcMsgError(extends OIDCError)
CryptoJWTError(extends OIDCError)
OidcOPError(extends OIDCError)
```

rohe:
Part of this is dealt with in #99 .

rohe:
If we add OidcError to oidcmsg it will not cover CryptoJWT but I don't see  that as a major problem.
Actually I think it's the right way to go.

peppelinux (approved):
@nsklikas overall It LGTM, so here my approval. Please fill evetually the remaining gaps, It seems that @rohe told us something ..

PR #99, opened 2021-06-12
Added a parent class exception OidcOPError.

rohe:
Moved a token error from exception.py to token/exception.py.
Used the umbrella error TokenException.


peppelinux (approved):
Good shot

nsklikas (approved):

PR #104, opened 2021-06-24
acr value in claims parameter was not treated correctly.

rohe:
If there was a 

{"id_token": {"acr": {"value": "https://refeds.org/profile/mfa"}}}

specification in the claims parameter in an authorization request the value was not treat correctly by the authn broker.


peppelinux (approved):
Not sure what it would be the else of some `if` but having seen the unit test there, looks good, great

PR #105, opened 2021-06-27
Allowing specification of a client DB in the configuration file.

rohe:
This is about allowing a specification of a client database in the configuration file.
And at the same make having that database being something persistent.


rohe:
Checks will fail until oidcmsg 1.4.0 has been released

peppelinux:
I approved oidcmsg PR, you can do release

PR #106, opened 2021-06-29
To fix token backward compatibility.

rohe:
This should fix the backward compatibility problem.


peppelinux:
It's definitely worth of It
Don't have any integration test for this, I approve but last Word to @nsklikas

peppelinux (approved):

nsklikas (approved):

PR #107, opened 2021-07-01
Userinfo claims

rohe:
When a specific acr is requested by using the claims parameter in an authorization request the resulting acr used should be reflected in the ID Token or at the user info endpoint as defined in the claims parameter.


rohe:
Don't know why it fails. Works flawlessly on my setup.
Will try to find out when I get back in a couple of hours.

peppelinux:
I read the code, seems clean and looks good.
Can't test by my side at this moment but I will asap

PR #108, opened 2021-07-02
Clear txt sid

rohe:
Session IDs in the old type tokens was in clear text. Not so in the new ones.
This takes care of the difference.


peppelinux (approved):
Code looks very good, I'll do my test this evening directly on the Dev branch, so let's merge this

rohe:
I'll wait to after lunch for @nsklikas to have his say.

PR #109, opened 2021-07-04
Version 2.1.0

rohe:
Fixed a couple of backward compatibility issues. Like accepting old tokens and allowing configuration of a client database separate from the session database.
Also made sure that if an acr value was specified using the claims parameter in authorization request, it would be honoured and reported back in the ID Token or in the response from the user info endpoint.


rohe:
Also made sure that if an acr value was specified using the claims parameter in authorization request, it would be honoured and reported back in the ID Token or in the response from the user info endpoint.

peppelinux:
Great, do we have to label as BREAKING CHANGE the new cdb approach?

Is there something to explain in the docs?

peppelinux:
I mean here
https://github.com/IdentityPython/oidc-op/pull/109/files#diff-0731523b7d97b481325cebda9df3e709982a684ddcaf77f73eeadd5e2a05c377R160

do we have to say that load/dump/flush methods MUST be adopted and this approach could be breaking for the implementers?

peppelinux (approved):
overall the code looks pretty ginvolvedood, great job Roland

rohe:
The new cdb approach isn't breaking anything. It allows new functionality that is what it does.
I rewrote AbstractFileSystem compared to previous usage to also have the dump/load methods. Not flush. Went back and forth on that but decided eventually not to support it.
If cdb is a class instance then the class must support dump and load.

peppelinux:
It looks very good to me, thank you

peppelinux:
tested on django-oidc-op, all passes, very good

nsklikas (approved):

PR #110, opened 2021-07-05
feat: added pypi gh action CD

peppelinux:
I made pypi release by hands, this action still missing.
I don't have access to settings, @rohe we just have to put there the right token


rohe:
I have put the token there.

PR #112, opened 2021-07-09
For debugging purpose nice to know what was put in the ID Token and a…

rohe:
…lso what was in a received ID Token.


peppelinux (approved):
Not sure if samesite lax or none, I should have to check if some POST is involved but I'm also sure that you made your tests (with chromium).

Good

rohe:
I'm  not sure either  regarding samesite. I just had to chose one or the other.

PR #113, opened 2021-07-27
WIP: Introduce add_claims_by_scope per client configuration

ctriant:
This MR enables per client configuration for add_claims_by_scope parameter.
If configuration for specific client exists, it overrides the global configuration.

For example consider the following list of clients and that the default oidc-op configuration sets `add_claims_by_scope = True`:

```yaml
oidc_clients:
  client1:
    # client secret is "password"
    client_secret: "Namnam"
    redirect_uris:
      - ['https://openidconnect.net/callback', '']
    response_types:
      - code
    add_claims_by_scope: False
    
  client2:
    client_secret: "spraket"
    redirect_uris:
      - ['https://app1.example.net/foo', '']
      - ['https://app2.example.net/bar', '']
    response_types:
      - code
```
Also consider the two clients form the following authorization request where `CLIENT_ID` is "client1" and "client2" respectively:

```
AREQS = AuthorizationRequest(
    response_type="code",
    client_id=CLIENT_ID,
    redirect_uri="http://example.com/authz",
    scope=["openid", "address", "email"],
    state="state000",
    nonce="nonce",
)

```
As long as for "client1"  `add_claims_by_scope = False`, the returned claims will be the union of the base claims and the explicit claims requested for this user given `enable_claims_per_client = True`.

On the other hand, for "client2" the returned claims will be the same as for "client1" plus the the claims that correspond to the requested scopes after `oidcop.scopes.convert_scopes2claims()` is applied.

### DIFF
--- a/src/oidcop/session/claims.py
+++ b/src/oidcop/session/claims.py
@@ -41,7 +41,11 @@ class ClaimsInterface:
 
     def _get_client_claims(self, client_id, usage):
         client_info = self.server_get("endpoint_context").cdb.get(client_id, {})
-        client_claims = client_info.get("{}_claims".format(usage), {})
+        client_claims = (
+            client_info.get("add_claims", {})
+            .get("always", {})
+            .get(usage, {})
+        )
         if isinstance(client_claims, list):
             client_claims = {k: None for k in client_claims}
         return client_claims
@@ -94,8 +98,19 @@ class ClaimsInterface:
 
         claims.update(base_claims)
 
-        # Scopes can in some cases equate to set of claims, is that used here ?
-        if module.kwargs.get("add_claims_by_scope"):
+        # If specific client configuration exists overwrite add_claims_by_scope
+        if client_id in _context.cdb:
+            add_claims_by_scope = (
+                _context.cdb[client_id].get("add_claims", {})
+                .get("by_scope", {})
+                .get(claims_release_point, {})
+            )
+            if isinstance(add_claims_by_scope, dict) and not add_claims_by_scope:
+                add_claims_by_scope = module.kwargs.get("add_claims_by_scope")
+        else:
+            add_claims_by_scope = module.kwargs.get("add_claims_by_scope")
+
+        if add_claims_by_scope:
             if scopes:
                 _scopes = _context.scopes_handler.filter_scopes(client_id, _context, scopes)
 
--- a/tests/test_01_claims.py
+++ b/tests/test_01_claims.py
@@ -114,6 +114,9 @@ class TestEndpoint(object):
             "client_salt": "salted",
             "token_endpoint_auth_method": "client_secret_post",
             "response_types": ["code", "token", "code id_token", "id_token"],
+            "add_claims": {
+                "always": {},
+            },
         }
         server.endpoint_context.keyjar.add_symmetric(
             "client_1", "hemligtochintekort", ["sig", "enc"]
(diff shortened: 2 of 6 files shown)


src/oidcop/session/claims.py
```diff
@@ -95,6 +95,10 @@ def get_claims(self, session_id: str, scopes: str, claims_release_point: str) ->
         claims.update(base_claims)
 
         # Scopes can in some cases equate to set of claims, is that used here ?
+        if not _context.cdb[client_id].get("add_claims_by_scope") is None:
+            module.kwargs["add_claims_by_scope"] = \
+                _context.cdb[client_id].get("add_claims_by_scope")
+
         if module.kwargs.get("add_claims_by_scope"):
```
nsklikas, line 102: You shouldn't overwrite the modules kwargs, this will affect later calls.

I think this would be cleaner:
```python
if "add_claims_by_scope" in _context.cdb[client_id]:
    add_claims_by_scope = _context.cdb[client_id]["add_claims_by_scope"]
else:
    add_claims_by_scope = module.kwargs.get("add_claims_by_scope"):

if add_claims_by_scope:
    ...
```

nsklikas (changes requested):

src/oidcop/session/claims.py
peppelinux: yes "if not is None" is heavy

peppelinux:
Thank you @ctriant 
can you rebase this PR to develop branch?
we cannot merge contributions directly on master branch

nsklikas (approved):

peppelinux:
unit tests fails, some changes are required

ctriant:
> unit tests fails, some changes are required

On it

src/oidcop/session/claims.py
peppelinux: Ok, done, great!

peppelinux:
Ok I'd merge It now but I'm wondering that It needs to be documented, otherwise only the code would show this feature.

@ctriant can you put a brief note on this feature, how It works with a short example?

ctriant:
> Ok I'd merge It now but I'm wondering that It needs to be documented, otherwise only the code would show this feature.
> 
> @ctriant can you put a brief note on this feature, how It works with a short example?

Hey @peppelinux, please see if the example on the MR description helps. Thanks!

peppelinux:
@criant yes it helps
Would you like to put it also in the documentation, where I found that we still don't have the "client" section documented!

https://github.com/IdentityPython/oidc-op/blob/master/docs/source/contents/conf.rst

@rohe ^

nsklikas:
Shouldn't this be per client? E.g introduce the options `userinfo_add_claims_by_scope`, `idtoken_add_claims_by_scope`, etc

rohe:
@nsklikas Since we have the client parameters `id_token_claims` and `userinfo_claims` it sounds reasonable to have `userinfo_add_claims_by_scope` and `idtoken_add_claims_by_scope`. Even though the long names makes me cringe.

Also, we might want to think about structuring this is some way.
That is instead of having:

```
oidc_clients = {
    "client_1": {
        "client_secret": "Namnam",
        "redirect_uris": [('https://openidconnect.net/callback', '')],
        "response_types": ["code"],
        "userinfo_add_claims_by_scope": False,
        "id_token_add_claims_by_scope": False,
        "userinfo_claims": ["email", "email_verified"],
        "id_token_claims": ["email", "phone"]
    }
}

```

we could have (to improve readability)

```

oidc_clients_s = {
    "client_1": {
        "client_secret": "Namnam",
        "redirect_uris": [('https://openidconnect.net/callback', '')],
        "response_types": ["code"],
        "add_claims": {
            "by_scope": {
                "userinfo": False,
                "id_token": False
            },
            "always": {
                "userinfo": ["email", "email_verified"],
                "id_token": ["email", "phone"]
            }
        }
    }
}

```

Just a thought.

peppelinux (approved):
@ctriant I just added some words in the documetation, the place I meant to put some note about this new feature

ctriant:
@peppelinux i think we better revert this MR because it was a WIP. 
Some tests are missing and also we have to make a decision regarding the configuration pattern before we are ready to merge ti.

peppelinux:
I should have known during the last developer meeting and this PR should have been a draft. How long does it take to complete this feature?

we will make a release solely to cover this issue, I beg you to give priority to this

ctriant:
The MR was marked as WIP so I thought this was enough to be considered as draft. Sorry for the inconvenience @peppelinux .

peppelinux:
don't worry, for the next time create drafts and non-PRs and switch to PR when ready for final revisions.

what time do you have to finish this WiP?

ctriant:
The implementation follows the proposed scheme by @rohe . If we decide not to adopt this proposal I have to revert some changes. Then add some tests. I think there is no much work needed to close it.

peppelinux:
yes I saw unit tests and I pushed an example in the documentation, it works as it is, please add your minor changes and unit test for the next, imminent, release and thank you for all of this

PR #115, opened 2021-08-04
Add pkce essential per client

nsklikas:
This allows us to have pkce(essential) activated for only some clients

### DIFF
--- a/src/oidcop/oidc/add_on/pkce.py
+++ b/src/oidcop/oidc/add_on/pkce.py
@@ -3,11 +3,9 @@ import logging
 from typing import Dict
 
 from cryptojwt.utils import b64e
-from oidcmsg.oauth2 import (
-    AuthorizationErrorResponse,
-    RefreshAccessTokenRequest,
-    TokenExchangeRequest,
-)
+from oidcmsg.oauth2 import AuthorizationErrorResponse
+from oidcmsg.oauth2 import RefreshAccessTokenRequest
+from oidcmsg.oauth2 import TokenExchangeRequest
 from oidcmsg.oidc import TokenErrorResponse
 
 from oidcop.endpoint import Endpoint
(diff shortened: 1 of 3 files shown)


peppelinux:
It looks good, I Just want to ask to you to put a brief description of this feature in the docs

Probably this Is good Moment to review the general policy of oidc-op with pkce, before merging this

nsklikas:
Where do you think is the right place to document this?

peppelinux:
https://github.com/IdentityPython/oidc-op/blob/master/docs/source/contents/conf.rst#add_on

What do you think to have a separate section for add_ons?

nsklikas:
Sure

nsklikas:
Do you think something like this https://github.com/nsklikas/oidc-op/blob/pkce-per-client/docs/source/contents/conf.rst ok?

src/oidcop/oidc/add_on/pkce.py
```diff
@@ -41,7 +39,12 @@ def post_authn_parse(request, client_id, endpoint_context, **kwargs):
     :param kwargs:
     :return:
     """
-    if endpoint_context.args["pkce"]["essential"] and "code_challenge" not in request:
+    client = endpoint_context.cdb[client_id]
+    if "pkce_essential" in client:
+        essential = client["pkce_essential"]
+    else:
+        essential = endpoint_context.args["pkce"]["essential"]
```
peppelinux, line 46: `essential = endpoint_context.args["pkce"].get("essential", False)`

@nsklikas how does it sounds to you?

peppelinux (approved):

src/oidcop/oidc/add_on/pkce.py
nsklikas: It's set in https://github.com/IdentityPython/oidc-op/pull/115/files#diff-60c1d62c690192be5964946a0883424948797ac8682e8ef94d9a59a6c8414d82L135. So it will always be there.

Should I remove that line and add the `get`?

peppelinux: Not so important neither urgent, only if it sounds good to you.
Yes explicit is better than implicit but an absent value (boolan) could be False by default.
That's my tastes, not mandatory for this project!

rohe:
I guess the 

`essential = essential = endpoint_context.args["pkce"].get(`

line is a typo ?
Shouldn't it be just

`essential = endpoint_context.args["pkce"].get(`

nsklikas:
Good catch.

peppelinux:
@rohe do we think that this could be merged as It Is or do we have to put something more in?

rohe:
I think we can go ahead with this as it is.

PR #116, opened 2021-08-04
Fix refresh tokens

nsklikas:
Fixes various issues with refresh:
- The id tokens issued after a refresh were not JWTs.
- Also fixed the way claims were added to the ID tokens, the claims that were passed from the grant were ignored.
- The scopes returned were not correct (we returned the scopes of the original authorization request, not of this request).
- We didn't always return an error when invalid scopes were requested

### DIFF
--- a/src/oidcop/oauth2/token.py
+++ b/src/oidcop/oauth2/token.py
@@ -253,7 +253,7 @@ class RefreshTokenHelper(TokenEndpointHelper):
         _resp = {
             "access_token": access_token.value,
             "token_type": access_token.token_type,
-            "scope": _grant.scope,
+            "scope": scope,
         }
 
         if access_token.expires_at:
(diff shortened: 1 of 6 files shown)


peppelinux:
@nsklikas for this chore I'd wait Roland before doing my revision

nsklikas:
I also pushed some minor fixes for refresh token:
- The scopes returned were not correct (we returned the scopes of the original authorization request, not of this request).
- We didn't always return an error when invalid scopes were requested

peppelinux:
> * The scopes returned were not correct (we returned the scopes of the original authorization request, not of this request).

Probably we had an issue on this problem, thank you, probably It wasn't fixed completely


> * We didn't always return an error when invalid scopes were requested

On invalid scoped we had an option in the old oidcendpoint to handle this, we should check

rohe (approved):

peppelinux (approved):

PR #117, opened 2021-08-16
Fix JWT access token lifetime

nsklikas:
The `expires_in` defined in usage rules was not reflected to the access token when it was a JWT

### DIFF
--- a/src/oidcop/token/id_token.py
+++ b/src/oidcop/token/id_token.py
@@ -267,6 +267,7 @@ class IDToken(Token):
         encrypt=False,
         code=None,
         access_token=None,
+        usage_rules: Optional[dict] = None,
         **kwargs,
     ) -> str:
         _context = self.server_get("endpoint_context")
--- a/src/oidcop/token/jwt_token.py
+++ b/src/oidcop/token/jwt_token.py
@@ -48,8 +48,13 @@ class JWTToken(Token):
         # inherit me and do your things here
         return payload
 
-    def __call__(self, session_id: Optional[str] = "", token_class: Optional[str] = "",
-                 **payload) -> str:
+    def __call__(
+        self,
+        session_id: Optional[str] = "",
+        token_class: Optional[str] = "",
+        usage_rules: Optional[dict] = None,
+        **payload
+    ) -> str:
 
         """
         Return a token.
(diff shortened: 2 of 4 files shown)


src/oidcop/token/id_token.py
```diff
@@ -249,7 +249,13 @@ def sign_encrypt(
 
         return _jwt.pack(_payload, recv=client_id)
 
-    def __call__(self, session_id: Optional[str] = "", ttype: Optional[str] = "", **kwargs) -> str:
+    def __call__(
+        self,
+        session_id: Optional[str] = "",
+        ttype: Optional[str] = "",
+        usage_rules: Optional[dict] = None,
```
peppelinux, line 256: None or {}?

it's a dict, probably {} Is better

src/oidcop/token/jwt_token.py
```diff
@@ -48,8 +48,13 @@ def load_custom_claims(self, payload: dict = None):
         # inherit me and do your things here
         return payload
 
-    def __call__(self, session_id: Optional[str] = "", token_class: Optional[str] = "",
-                 **payload) -> str:
+    def __call__(
+        self,
+        session_id: Optional[str] = "",
+        token_class: Optional[str] = "",
+        usage_rules: Optional[dict] = None,
```
peppelinux, line 55: None or {}?

it's a dict, probably {} Is better

peppelinux (approved):

src/oidcop/token/id_token.py
rohe: I've always (well almost) refrained from using mutable values as defaults. So I'd vote for None.

rohe (approved):

src/oidcop/token/id_token.py
nsklikas: Yes, a dict for a default value is a bug waiting to happen

peppelinux: I've always thought about why a variable of one type could take on another type, such as None. But it's a completely cosmetic thing in python, let's move on

peppelinux:
@nsklikas we just have to patch the conflicting files, then merge

peppelinux (approved):

PR #118, opened 2021-08-16
Handle exception raised from expired JWT access token

nsklikas:
Using an expired JWT access token in the userinfo endpoint would raise an unhandled `ToOld` exception. This PR fixes that

### DIFF
--- a/src/oidcop/client_authn.py
+++ b/src/oidcop/client_authn.py
@@ -22,6 +22,7 @@ from oidcop.endpoint_context import EndpointContext
 from oidcop.exception import InvalidClient
 from oidcop.exception import MultipleUsage
 from oidcop.exception import NotForMe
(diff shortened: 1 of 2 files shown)


src/oidcop/client_authn.py
```diff
@@ -22,6 +22,7 @@
 from oidcop.exception import InvalidClient
 from oidcop.exception import MultipleUsage
 from oidcop.exception import NotForMe
+from oidcop.exception import ToOld
```
peppelinux, line 25: Does It mean tooOld?

peppelinux (approved):

src/oidcop/client_authn.py
rohe: No, ToOld is correct.

rohe (approved):

PR #119, opened 2021-08-19
Missing userinfo in Configuration now raises an exception

peppelinux:
Working on SATOSA integration I found that when userinfo is missign in oidcop configuration we get

`TypeError: 'NoneType' object is not callable`

With this PR we get instead a more eloquent exception (message)


peppelinux:
ops, I'm rebasing it to developr branch... sorry!

PR #123, opened 2021-09-06
Compare client_id against grant's client

nsklikas:
A client could take a token from another client and use it as his own.

### DIFF
--- a/src/oidcop/oauth2/token.py
+++ b/src/oidcop/oauth2/token.py
@@ -198,6 +198,11 @@ class AccessTokenHelper(TokenEndpointHelper):
             return self.error_cls(error="invalid_grant", error_description="Unknown code")
 
         grant = _session_info["grant"]
+        if grant.authorization_request["client_id"] != client_id:
+            return self.error_cls(
+                error="invalid_client", error_description="Invalid client"
+            )
+
         code = grant.get_token(request["code"])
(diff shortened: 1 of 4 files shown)


peppelinux (approved):

rohe:
So, this was one of the things I fixed in the code as a result of the conformance testing.
Different solution from your proposal though.
Have to compare.

peppelinux:
in django-oidc-op and the satosa frontend based in oidcop, due to their storage logic based on flush/dump/load/ method of session_manager, this problem is fixed by design. Here the test

https://github.com/peppelinux/SATOSA/commit/df6e9a880394422ef84afb4b672d6d28b0223e43#diff-f10182b97d02a823560e9df6c[…]778db6a2f6050bcc03337c3d67R765

anyway this PR is very important for who use oidcop with a pure inmemory storage or another inherited engine from inmemory interfaces, good job @nsklikas

rohe:
There are 2 differences between @nsklikas proposal and mine.
I picked client_id from the session_info so my check happens a couple of lines earlier then @nsklikas's 
I use the error code invalid_grant which explicitly lists 'or was issued to another client.' as the cause of the error.

I've added the new tests to my branch since I think we should choose that one instead of this. My 2 cents.

nsklikas:
Your solution sounds better. I will close this PR when you open one with your changes.

PR #125, opened 2021-09-07
Certification testing - bug fixes

rohe:
Bugs/wrong behaviour found doing certification testing:

- Client registration endpoint should return a 201 on successful client registration
- Default token lifetime set to 30 minutes (was none)
- Userinfo endpoint should support POST
- Authorization error response MUST contain 'state' if present in request
- Cookies was not picked properly - looked for wrong name. 
- Cookies was picked twice. Supposed to be only once
- Honor max_age set in authentication request
- invalid_grant instead of invalid_request when wrong token was used.
- Revoke all tokens minted based on an authorization_code on seconds usage of the authorization_code.
- Verify that the user of a token was actually the owner of the same.
- prompt=login in authentication request SHOULD force re-authentication.


rohe:
I'll fix the build errors.
But not until  later today. Have to do other stuff for a while.

peppelinux:
I would like to ask you to rebase all those duplicated commits, but don't worry, I won't :smile_cat:

nsklikas (approved):

PR #128, opened 2021-09-10
Don't issue refresh token if not configured

nsklikas:
This should have been in my last PR, we should respect the configured grant_types and not issue a refresh token if the `refresh_token` grant type is not supported

### DIFF
--- a/src/oidcop/oidc/token.py
+++ b/src/oidcop/oidc/token.py
@@ -45,11 +45,16 @@ class AccessTokenHelper(TokenEndpointHelper):
         _session_info = _mngr.get_session_info_by_token(_access_code, grant=True)
         logger.debug(f"Session info: {_session_info}")
 
-        if _session_info["client_id"] != req["client_id"]:
-            logger.debug("{} owner of token".format(_session_info["client_id"]))
+        client_id = _session_info["client_id"]
+        if client_id != req["client_id"]:
+            logger.debug("{} owner of token".format(client_id))
             logger.warning("{} using token it was not given".format(req["client_id"]))
             return self.error_cls(error="invalid_grant", error_description="Wrong client")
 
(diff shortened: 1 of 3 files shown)


peppelinux:
It seems working good, can you fix the conflicting files?

@rohe that's for the next major realease with all the contributions in, or do we think to merge only this pr for an intermediate release?

src/oidcop/oidc/token.py
```diff
@@ -45,11 +45,16 @@ def process_request(self, req: Union[Message, dict], **kwargs):
         _session_info = _mngr.get_session_info_by_token(_access_code, grant=True)
         logger.debug(f"Session info: {_session_info}")
 
-        if _session_info["client_id"] != req["client_id"]:
-            logger.debug("{} owner of token".format(_session_info["client_id"]))
+        client_id = _session_info["client_id"]
```
peppelinux, line 48: that's the same code of before, in oauth2/token.py ...


Mhh, can we have a decoupled function that returns a boolean (true it_can, false no_it_cannot ) for doing this check?

peppelinux (changes requested):

rohe:
I have a couple of changes based on the logout testing.
Stalled because I found an error in the test suite and they are slow to fixing.
Anyway, they could hold for 2.3.0 (I think 2.3.0 is more appropriate then 2.2.1)

src/oidcop/oidc/token.py
nsklikas: There is a lot of duplicate code between these two files.

We probably need a small refactor.

peppelinux: if you can do it in this PR it would be better, every left behind becomes then forgotten :)

peppelinux (approved):

rohe (approved):

PR #129, opened 2021-09-10
Add client configuration documentation

nsklikas:
We should document any client configuration available.

I will try to gradually add any configuration i find.

Feel free to help if anyone wants @ctriant @peppelinux


peppelinux:
@nsklikas following the latest changes can you resolve those conflicts in conf.rst?

Before the next release we should do a reconnaissance of all critical changes made in the last commits and evaluate the coverage within the documentation

peppelinux:
@rohe do we agree that this PR could be merged as it is and then a new release of oidcop would see the light?

rohe:
Sure, go ahead!

peppelinux:
@nsklikas is there something more that you want to add to this pr?
we know that there's many thing to be improved and added to the documentation... anyway let's have a new release for today if you agree

PR #130, opened 2021-09-13
Refactor scopes configuration

### LINKED ISSUE #127: Scopes handling
I will use this issue to try to describe how I think we should handle scopes. When we agree to something I will go forward and open a PR.

Right now we have `scope_supported` which is used to define the scopes that are advertised in the well-known endpoint as well as the scopes that we allow. This can be overridden per client with the `allowed_scopes` parameter.
If someone wants to override the scopes mapping then he must use the custom_scopes add on.

We would like to be able not to advertise all the scopes that are supported. Also I think that it would be nice to have a scopes mapping per client.
So I propose that remove the existing scopes configuration and introduce the following:
-  `scopes_mapping` (maybe simply `scopes` is a better name?): A dict defining the scopes that are allowed to be used per client and the claims they map to (defaults to the scopes mapping described in the spec). If we want to define a scope that doesn't map to claims (e.g. offline_access) then we simply map it to an empty list. E.g.:
```json
{
  "scope_a": ["claim1", "claim2"],
  "scope_b": []
}
```
-  `scopes_supported` (maybe `allowed_scopes` is better?): A list with the scopes that are allowed to be used (defaults to the keys in `scopes_mapping`).
- `advertised_scopes`: A list with the scopes that will be advertised in the well-known endpoint (defaults to `scopes_supported`).

In addition to those configuration options we should also have the corresponding per client ones for `scopes_supported` and `scopes_mapping`.

@rohe @peppelinux what do you think? Should I go ahead and implement this?

nsklikas:
I went ahead and implemented the behavior described in https://github.com/IdentityPython/oidc-op/issues/127. I hope you agree with my approach.

If this PR is merged then the `custom_scopes` add on should be deprecated. I didn't delete it in case someone uses it in his installation but added a warning.

Closes #127

### DIFF
--- a/docs/source/contents/conf.rst
+++ b/docs/source/contents/conf.rst
@@ -53,6 +53,28 @@ sub_funcs
 
 Optional. Functions involved in *sub*ject value creation.
 
+
+
+scopes_mapping
+##############
+
+A dict defining the scopes that are allowed to be used per client and the claims
+they map to (defaults to the scopes mapping described in the spec). If we want
+to define a scope that doesn't map to claims (e.g. offline_access) then we
+simply map it to an empty list. E.g.::
+  {
+    "scope_a": ["claim1", "claim2"],
+    "scope_b": []
+  }
+*Note*: For OIDC the `openid` scope must be present in this mapping.
+
+
+allowed_scopes
+##############
+
+A list with the scopes that are allowed to be used (defaults to the keys in scopes_mapping).
+
+
 ------
 add_on
 ------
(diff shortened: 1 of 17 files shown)


rohe:
Sorry, should have reply long time ago. Was busy elsewhere.
I'm OK with your proposal. I don't see any obvious problems with it.

rohe (approved):

peppelinux (approved):
Thanks to Roland that triggers a message to me automatically, I was fare from this thread for days ...

Yes, I appreciated the issue and this pr as well

PR #132, opened 2021-09-21
Catch unhandled exception

nsklikas:
`get_session_info_by_token` might throw an exception, we catch it and return the appropriate error.


peppelinux (approved):
Thanks, really good

rohe (approved):

PR #137, opened 2021-09-22
Add parameter to revoke old refresh token upon issuing new

ctriant:
Introduce parameter `revoke_refresh_on_issue` in order to revoke the Refresh Token used to issue a new one.

### DIFF
--- a/src/oidcop/oidc/token.py
+++ b/src/oidcop/oidc/token.py
@@ -81,12 +81,10 @@ class AccessTokenHelper(TokenEndpointHelper):
 
         logger.debug("All checks OK")
 
-        issue_refresh = False
-        if "issue_refresh" in kwargs:
-            issue_refresh = kwargs["issue_refresh"]
-        else:
-            if "offline_access" in grant.scope:
-                issue_refresh = True
+        issue_refresh = kwargs.get("issue_refresh", None)
+        # The existence of offline_access scope overwrites issue_refresh
+        if issue_refresh is None and "offline_access" in grant.scope:
+            issue_refresh = True
 
         _response = {
             "token_type": token_type,
@@ -242,12 +240,12 @@ class RefreshTokenHelper(TokenEndpointHelper):
             _resp["expires_in"] = access_token.expires_at - utc_time_sans_frac()
 
         _mints = token.usage_rules.get("supports_minting")
-        issue_refresh = False
-        if "issue_refresh" in kwargs:
-            issue_refresh = kwargs["issue_refresh"]
-        else:
-            if "offline_access" in scope:
-                issue_refresh = True
+
+        issue_refresh = kwargs.get("issue_refresh", None)
+        # The existence of offline_access scope overwrites issue_refresh
+        if issue_refresh is None and "offline_access" in scope:
+            issue_refresh = True
+
         if "refresh_token" in _mints and issue_refresh:
             refresh_token = self._mint_token(
                 token_class="refresh_token",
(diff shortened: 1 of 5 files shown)


src/oidcop/oidc/token.py
```diff
@@ -280,6 +277,8 @@ def process_request(self, req: Union[Message, dict], **kwargs):
             _resp["id_token"] = _idtoken.value
 
         token.register_usage()
+        if kwargs.get("revoke_refresh_on_issue", False):
```
nsklikas, line 280: I think that this should be on the class (for global configuration) and on the client (for per-client configuration). We shouldn't have to pass it to the function every time. Maybe keep this and add the other 2 methods?

Also I think `True` is a better default value.

src/oidcop/oidc/token.py
```diff
@@ -242,12 +239,12 @@ def process_request(self, req: Union[Message, dict], **kwargs):
...
-        if "issue_refresh" in kwargs:
-            issue_refresh = kwargs["issue_refresh"]
-        else:
-            if "offline_access" in scope:
-                issue_refresh = True
+
+        issue_refresh = kwargs.get("refresh_token", False)
+        # The existence of offline_access scope overwrites issue_refresh
+        if "offline_access" in scope:
+            issue_refresh = True
```
nsklikas, line 246: The logic is reversed here. `issue_refresh` used to override `offline_access`, which IMHO is the correct behavior. Maybe this is better:
```python
        issue_refresh = kwargs.get("refresh_token", None)
        # The existence of offline_access scope overwrites issue_refresh
        if issue_refresh is None and "offline_access" in scope:
            issue_refresh = True
```

rohe (approved):

src/oidcop/oidc/token.py
```diff
@@ -81,12 +81,9 @@ def process_request(self, req: Union[Message, dict], **kwargs):
...
 
-        issue_refresh = False
-        if "issue_refresh" in kwargs:
-            issue_refresh = kwargs["issue_refresh"]
-        else:
-            if "offline_access" in grant.scope:
-                issue_refresh = True
+        issue_refresh = kwargs.get("issue_refresh", False)
+        if "offline_access" in grant.scope:
+            issue_refresh = True
```
nsklikas, line 86: The logic is reversed here as well. `issue_refresh` used to override `offline_access`, which IMHO is the correct behavior. Maybe this is better:
```python
        issue_refresh = kwargs.get("refresh_token", None)
        # The existence of offline_access scope overwrites issue_refresh
        if issue_refresh is None and "offline_access" in scope:
            issue_refresh = True
```

src/oidcop/oidc/token.py
```diff
@@ -281,6 +278,19 @@ def process_request(self, req: Union[Message, dict], **kwargs):
...
         token.register_usage()
 
+        if ("client_id" in req
+            and req["client_id"] in _context.cdb
+            and "revoke_refresh_on_issue" in _context.cdb[req["client_id"]]
+        ):
+            revoke_refresh = _context.cdb[req["client_id"]].get("revoke_refresh_on_issue")
+        else:
+            revoke_refresh =(
+                _mngr.conf["token_handler_args"]["refresh"]["kwargs"].get("revoke_refresh_on_issue", False)
```
nsklikas, line 288: Why add it to the token handler? Do you think it's a better there than the token class? You could add it to the token endpoint (oidcop/{oauth2,oidc}/token.py:Token) and then we could have here:
```python
revoke_refresh = self.endpoint.revoke_refresh_on_issue
```
what do think about this?

ctriant: I thought token handler as a more suitable place as it describes the behavior upon tokens, in contrast with token class that describe the actual token.

The truth is that in oauth2/token.py: https://github.com/IdentityPython/oidc-op/blob/66eb57b305258308365b5439bb1f41c32950018b/src/oidcop/oauth2/token.py#L346 there is an unused parameter `new_refresh_token` that it seems to resembles the same functionality as `revoke_refresh`.

So I have no reason to insist on this decision.

peppelinux (approved):
I like this feature, thank you guys!
@ctriant let us know when this PR would be ready to be merged, we're watching many integrations, take your time and give us a know when ready for merge

ctriant:
> I like this feature, thank you guys! @ctriant let us know when this PR would be ready to be merged, we're watching many integrations, take your time and give us a know when ready for merge

@peppelinux i think we are ready, i integrated the suggestions of @nsklikas

src/oidcop/oidc/token.py
peppelinux: ok @nsklikas I wait for your revision before merge

nsklikas (approved):

Issue #142, opened 2021-10-18
documentation inconsistency


melanger:
In many tests and examples, including the [main example](https://github.com/IdentityPython/oidc-op/blob/a87177e7a5d0b40dcff11ed75b39a72bd2d3f015/docs/source/contents/conf.rst), the configuration option for claims is written as `add_claim_by_scope` (singular `claim`), but in the source code, it is `add_claims_by_scope` (plular `claims`):

https://github.com/IdentityPython/oidc-op/blob/a87177e7a5d0b40dcff11ed75b39a72bd2d3f015/src/oidcop/session/claims.py#L28

https://github.com/IdentityPython/oidc-op/blob/a87177e7a5d0b40dcff11ed75b39a72bd2d3f015/src/oidcop/oidc/userinfo.py#L38

I guess the documentation and tests should be fixed to `add_claims_by_scope`.

peppelinux:
Think you're right
@nsklikas can you confirm this?

@melanger feel free to put your contribution with a Pull Request, thank you

PR #143, opened 2021-10-19
fix Add claims by scope configuration option name

melanger:


peppelinux:
This closes https://github.com/IdentityPython/oidc-op/issues/142

PR #144, opened 2021-10-20
Certification 2

rohe:
Contains the following changes

- Client registration endpoint should return a 201 HTTP response on successful client registration.
- Default token lifetime should not be 0 (zero=infinite). Changed to be 30 minutes.
- Authorization error response MUST contain ‘state’ if it is present in the request.
- Made sure cookies was only parsed once.
- To old cookie should result in re-authentication.
- Fixed incorrect error code on usage of unusable ‘authorization code’.
- Revoke tokens that has been minted using a code that then is used once more.
- Make sure a refresh token is used by the client that it was given too.
- prompt==login forces re-authentication.
- ‘prompt’ value is a list.
- If response_type==id_token return userinfo that normally would be returned from the userinfo endpoint in the id_token.
- form_post results in no return_uri or response_args in resp_info.
- Error response when response_mode==form_post not handled correctly. return_uri needed.
- Use extra info in ValueError exception instance to return better error code.
- Allow endpoints to have their own request verification error handler - added method.
- Having session ID in ID Token instance simplifies things.


.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,29 @@
+{
```
peppelinux, line 1: .pytest_cache may be ignored in .gitignore and removed from this PR

debug/authreq.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python3
```
peppelinux, line 1: @rohe is `debug/` folder relevant for the project?

setup.py
```diff
@@ -72,7 +72,7 @@ def run_tests(self):
         "Programming Language :: Python :: 3.9",
         "Topic :: Software Development :: Libraries :: Python Modules"],
     install_requires=[
-        "oidcmsg==1.4.0",
+        "oidcmsg==1.4.1",
```
peppelinux, line 75: oidcmsg==1.4.4 here I presume

example/flask_op/certs/89296913_127.0.0.1.cert
```diff
@@ -0,0 +1,18 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
rohe: We should in the example directory.

unsupported/chpy/templates/user_pass.jinja2
```diff
@@ -0,0 +1,39 @@
+<!doctype html>
```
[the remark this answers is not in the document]
rohe: I'll remove it.

rohe:
Mismatch between @nsklikas changes to claims handling and what I've got.
Have to close this PR and try to merge the two branches.

PR #147, opened 2021-10-28
Update to last oidcmsg version

MdreW:


peppelinux:
Closed by
https://github.com/IdentityPython/oidc-op/pull/150

thank you @MdreW, with your help we found a huge dependency issue with oidcmsg and cryptojwt ... That's definitely fixed with oidcmsg v1.5.0

PR #148, opened 2021-11-01
docs: add prompt to refresh token requirements

melanger:
re https://github.com/UniversitaDellaCalabria/SATOSA-oidcop/issues/1

### DIFF
--- a/docs/source/contents/usage.md
+++ b/docs/source/contents/usage.md
@@ -52,8 +52,9 @@ We can even test the single logout
 Refresh token
 -------------
 
-Here an example about how to refresh a token.
-It is important to consider that only scope=offline_access will get a usable refresh token.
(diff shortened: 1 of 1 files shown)


peppelinux:
Good contribution, thank you!
@rohe we know that's something that belong to oidc certification requirements. @melanger may we have to say this in the docs?

I mean, that's not am oidcop requirement but a oidc specification one

rohe:
True, it's an OIDC specification requirement. Though there are caveats, as always. :-/

peppelinux:
Ok I think it's worth of it to specify this behaviour also in the documentation, specially for legacy clients and consumers that would think that oidcop has some strange behaviour :)

I would like just to underline that's something required from oidc certification, and not a free initiative of the oidcop developers :)

melanger:
> Ok I think it's worth of it to specify this behaviour also in the documentation, specially for legacy clients and consumers that would think that oidcop has some strange behaviour :)
> 
> I would like just to underline that's something required from oidc certification, and not a free initiative of the oidcop developers :)

Note added.

peppelinux (approved):

peppelinux:
@rohe this is an easy shot, once merged the official docs will be updated automatically (because it's on master branch)

rohe (approved):

PR #149, opened 2021-11-02
feat: add RawID to session manager

melanger:
allow passing already hashed IDs from backends or hashing in a microservice instead

### DIFF
--- a/src/oidcop/session/manager.py
+++ b/src/oidcop/session/manager.py
@@ -25,6 +25,14 @@ from ..token.handler import TokenHandler
 logger = logging.getLogger(__name__)
 
 
+class RawID(object):
+    def __init__(self, *args, **kwargs):
+        pass
+
+    def __call__(self, uid, *args, **kwargs):
+        return uid
+
+
 def pairwise_id(uid, sector_identifier, salt="", **kwargs):
     return hashlib.sha256(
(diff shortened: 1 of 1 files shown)


peppelinux:
Ciao @melanger 

unfortunately we have these subject types following the oidc official specs, here:
https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes

Consider to install your own python package with and configure it in your oidcop sub functions (you have them in the general configuration)

anyway I ask a revision to other friends that helps in releasing new versions of oidcop

melanger:
@peppelinux I don't mean it as a new subject type (such as `subject_type=raw`), but as an alternative implementation of public ID. Maybe it could be RawPublicID? But yes, I can put this into a separate module.

The reason for this proposal is that with PublicID and SubjectID from oidcop it is not possible to use anything else than a SHA256 hash of userID+section_identifier+salt. It is not possible to use a raw user attribute (e.g. I already have eduPersonUniqueID so I do not need to hash it again) or to do other hashing (e.g. use the Hasher microservice in SATOSA instead).

peppelinux:
@melanger I understand perfectly!
let's take some more days and wait for the revision of roland and nikos, there's no hurry

rohe (approved):

peppelinux (approved):

PR #153, opened 2021-11-14
Certification 2

rohe:
A set of changes needed to pass OIDC certification.
Mainly on form_post and the single logout side.


peppelinux:
Ohhh, 160 commits 🤣
Ok, let's do the revision ...

rohe:
Yeah, I wish I could compact the branch. Tried it a while ago didn't work.
If you have a good idea about how to accomplish it I'd like to hear it.
Because some of these 160 commits are just back and forth that in the end would even out.

rohe:
I'll close this PR and will try to compact it. Unfortunately it looks like it will result in not 160 commits but one huge one.

PR #155, opened 2021-11-15
OIDC certification 

peppelinux:
Latest rohe's improvements on OIDC certification 

- FAPI OP example project


example/flask_op/config.json
```diff
@@ -116,6 +116,21 @@
...
+          "ES256",
+          "ES384",
+          "ES512",
+          "HS256",
+          "HS384",
+          "HS512",
+          "PS256",
+          "PS384",
+          "PS512",
+          "none"
```
peppelinux, line 133: sure to include none in the example algs?

example/flask_op/config_martin.yaml
```diff
@@ -0,0 +1,309 @@
+logging:
```
peppelinux, line 1: is it needed? :)

setup.py
```diff
@@ -72,7 +72,7 @@ def run_tests(self):
         "Programming Language :: Python :: 3.9",
         "Topic :: Software Development :: Libraries :: Python Modules"],
     install_requires=[
-        "oidcmsg==1.4.0",
+        "oidcmsg==1.5.0",
```
peppelinux, line 75: finally, wow!

pyproject.toml
```diff
@@ -7,7 +7,7 @@ build-backend = "setuptools.build_meta"
 
 [metadata]
 name = "oidcop"
-version = "2.1.0"
+version = "2.3.0"
```
peppelinux, line 10: mmhhh... it's already in src/.__init__.py

peppelinux:
please @rohe I think that we should pruned

example/flask_op/templates.tgz

peppelinux:
Closes https://github.com/IdentityPython/oidc-op/pull/150

example/flask_op/config.json
rohe: Maybe not.

example/flask_op/config_martin.yaml
rohe: Not really.

setup.py
rohe: Latest oidcmsg version is 1.5.1 we should probably point to that.

pyproject.toml
rohe: Don't know which has precedence.

rohe (approved):

setup.py
peppelinux: may we have an additional commit in this PR for achieve this?

pyproject.toml
peppelinux: we need a strategy here to prevent to update the same value in two different place/files

https://pypi.org/project/poetry-dynamic-versioning/

That would be more like an issue to be implemented in the near future

rohe: Yes ! We're using poetry för CryptoJWT and it's frustrating moving between that framework and the one we
use for all the others. I think we should eventually move to poetry for all libraries.

setup.py
rohe: Yes! Will do.

example/flask_op/config.json
rohe: Removed

example/flask_op/config_martin.yaml
rohe: The config_martin.yaml file is removed. Not updated and found no compelling reason to do it. Removed it instead.

rohe:
Changes that this PR accomplish:

- If response_type == id_token return userinfo that normally would be returned from the userinfo endpoint in the id_token.
- Don't run parse_cookie twice. One after the other.
- Allow for different claims release policies to be in effect simultaneously (typically id_token+userinfo).
- Error response when response_mode == form_post not handled correctly. return_uri needed.
- Use extra info in ValueError exception instance to return correct error code.
- Allow endpoints to have their own request verification error handler.
- Having session ID in ID Token instance simplifies things.
- Finding the last issued token of a specific type is useful.
- Should check if client session is revoked.
- Revoked all grants issued to a user/client pair.
- Merged the two ClaimsInterface versions.
- Removed tests using method that was removed and incomplete test.
- Client registration endpoint should return a 201 HTTP response code on successful registration.
- Default token lifetime should not be 0 (zero). Changed to be 30 minutes (1800 seconds).
- Authorization error response MUST contain 'state' if it is present in the request.
- Check cookie age.
- Revoke tokens that has been minted using a code that then is used once more.
- prompt==login forces re-authentication.
- prompt attribute is a list.

peppelinux:
@rohe we have cryptojwt 1.6.0 in pypi
https://pypi.org/project/cryptojwt/

why oidcmsg doesn't work?

rohe:
Hmm, have to look into it.
I'll do that tomorrow.

setup.py
```diff
@@ -72,8 +72,7 @@ def run_tests(self):
         "Programming Language :: Python :: 3.9",
         "Topic :: Software Development :: Libraries :: Python Modules"],
     install_requires=[
-        "oidcmsg==1.5.0",
-        "cryptojwt==1.5.2",
```
peppelinux, line 76: yes, cryptojwt comes with oidcmsg

peppelinux:
yes CI fails because we need first of all the release oidcmsg 1.5.3, due to this
https://github.com/IdentityPython/JWTConnect-Python-OidcMsg/pull/53

tests/test_31_oauth2_introspection.py
```diff
@@ -225,7 +224,7 @@ def _mint_token(self, token_class, grant, session_id, based_on=None, **kwargs):
             endpoint_context=self.token_endpoint.server_get("endpoint_context"),
             token_class=token_class,
             token_handler=self.session_manager.token_handler.handler[token_class],
-            expires_at=time_sans_frac() + 300,  # 5 minutes from now
+            expires_at=utc_time_sans_frac() + 300,  # 5 minutes from now
```
peppelinux, line 227: @rohe this 300 must be defined in the oidcop configuration together with a timeslack for verifications

rohe: OK, I'll do it.

rohe: The proper name is maximum clock skew :-)

peppelinux:
ok, so hopefully we'll have first the next release of oidcmsg to get oidcop CI working again.
Probably we'll have in this PR also the configuration paramers for time slack

peppelinux:
@rohe oidcmsg 1.5.3 just released, this CI works great
it seems to me that this is ready to be merged

rohe:
I think so too. Waiting for Nikos!

src/oidcop/oauth2/token.py
```diff
@@ -477,7 +477,9 @@ def process_request(self, request: Optional[Union[Message, dict]] = None, **kwar
             name=_context.cookie_handler.name["session"],
             sub=_session_info["grant"].sub,
             sid=_context.session_manager.session_key(
-                _session_info["user_id"], _session_info["user_id"], _session_info["grant"].id,
+                _session_info["user_id"],
+                _session_info["user_id"],
```
nsklikas, line 481: Shouldn't this be `_session_info["client_id"]`?

rohe: yes

tests/test_100_temp.py
```diff
@@ -0,0 +1,422 @@
+import copy
```
nsklikas, line 1: This file should have a better name, what is this supposed to test?

rohe: It was/is a placeholder for tests to come.
We can remove it for now.

src/oidcop/oauth2/token.py
peppelinux: I'm going to fix this, thank you @nsklikas

tests/test_100_temp.py
peppelinux: I'll simply rename it

src/oidcop/oauth2/token.py
peppelinux: mmhh .. it's this way also in the current master branch!
https://github.com/IdentityPython/oidc-op/blob/05611f16a31ce9092b83b00888cebc61cf98ca55/src/oidcop/oauth2/token.py#L480

ok, consider it fixed

tests/test_100_temp.py
peppelinux: pruned here
https://github.com/IdentityPython/oidc-op/pull/155/commits/accc68d26c1add00adde6c474eb1cfe256b06e40

PR #163, opened 2021-12-02
Check if token exists

nsklikas:
The access token may be a revoked/deleted access token so we must not assume that it exists in the grant's issued tokens.


rohe (approved):
If we start deleting tokens that has been revoked/timed-out from the session DB then this really is needed.

peppelinux (approved):

PR #165, opened 2021-12-09
Token exchange support

ctriant:


src/oidcop/oidc/token.py
```diff
@@ -352,6 +362,284 @@ def post_parse_request(
...
+                error="invalid_request", error_description="Wrong token type"
+            )
+
+        if token.is_active() is False:
+            return self.error_cls(
+                error="invalid_request", error_description="Subject token inactive"
+            )
+        return request
+
+    def check_for_errors(self, request):
```
nsklikas, line 440: Perhaps a better name would be `enforce_policy`

src/oidcop/oidc/token.py
```diff
@@ -352,6 +362,284 @@ def post_parse_request(
...
+                error="invalid_request",
+                error_description="Subject token invalid"
+            )
+
+        token = _mngr.find_token(_session_info["session_id"], request["subject_token"])
+
+        if not isinstance(token, (AccessToken, RefreshToken)):
+            return self.error_cls(
+                error="invalid_request", error_description="Wrong token type"
+            )
```
nsklikas, line 432: Let's remove this, I think this should be done in `enforce_policy`/`check_for_errors`

src/oidcop/oidc/token.py
```diff
@@ -352,6 +362,284 @@ def post_parse_request(
...
+        response_args["access_token"] = access_token.value
+        response_args["scope"] = access_token.scope
+        if refresh_token is None:
+            response_args["issued_token_type"] = "urn:ietf:params:oauth:token-type:access_token"
+            response_args["token_type"] = access_token.token_type
+            response_args["expires_in"] = access_token.usage_rules["expires_in"]
+        else:
+            response_args["issued_token_type"] = "urn:ietf:params:oauth:token-type:refresh_token"
+            response_args["refresh_token"] = refresh_token.value
+            response_args["expires_in"] = refresh_token.usage_rules["expires_in"]
```
nsklikas, line 497: Let's not issue a refresh token for now

src/oidcop/oidc/token.py
```diff
@@ -352,6 +362,284 @@ def post_parse_request(
...
+
+        if not isinstance(token, (AccessToken, RefreshToken)):
+            return self.error_cls(
+                error="invalid_request", error_description="Wrong token type"
+            )
+
+        if token.is_active() is False:
+            return self.error_cls(
+                error="invalid_request", error_description="Subject token inactive"
+            )
```
nsklikas, line 539: This is done in `post_parse_request`

src/oidcop/oidc/token.py
```diff
@@ -352,6 +362,284 @@ def post_parse_request(
...
+                )
+            except MintingNotAllowed:
+                logger.error("Minting not allowed for 'refresh_token'")
+                return self.error_cls(
+                    error="invalid_grant",
+                    error_description="Token Exchange not allowed with that token",
+                )
+
+            return self.token_exchange_response(access_token=new_token,
+                                                refresh_token=refresh_token)
```
nsklikas, line 642: No need for separate logic here

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,249 @@ def post_parse_request(
...
+            )
+        return resp
+
+    def enforce_policy(self, request, token):
+        _context = self.endpoint.server_get("endpoint_context")
+
+        if not isinstance(token, (AccessToken, RefreshToken)):
+            return self.error_cls(
+                error="invalid_request", error_description="Wrong token type"
+            )
```
nsklikas, line 443: This should be removed, we check this with `subject_token_types_supported`

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,249 @@ def post_parse_request(
...
+
+        if (
+            "requested_token_type" in request
+            and request["requested_token_type"] not in self.config["token_exchange"]["requested_token_types_supported"]
+        ):
+            return TokenErrorResponse(
+                error="invalid_request",
+                error_description="Unsupported requested token type",
+            )
```
nsklikas, line 458: I'm not sure if this should exist, can you post an example configuration in the issue so we can discuss it there?

My problem is that depending on the token type we may wish to support different `requested_token_types`

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,249 @@ def post_parse_request(
...
+
+        elif (
+            "requested_token_type" in request
+            and request["requested_token_type"] == "urn:ietf:params:oauth:token-type:refresh_token"
+            and request["subject_token_type"] != "urn:ietf:params:oauth:token-type:refresh_token"
+        ):
+            return TokenErrorResponse(
+                    error="invalid_request",
+                    error_description=f"Exchange {request['subject_token_type']} to refresh token forbbiden",
+                )
```
nsklikas, line 479: Do we need to enforce this? Shouldn't this be configurable?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,249 @@ def post_parse_request(
...
+
+        if (
+            "requested_token_type" in request
+            and request["requested_token_type"] in self.config["token_exchange"]["policy"]
+        ):
+            fun = importer(self.config["token_exchange"]["policy"][request["requested_token_type"]]["callable"])
```
nsklikas, line 485: Perhaps `fn` would be a better name?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,249 @@ def post_parse_request(
...
+            kwargs = self.config["token_exchange"]["policy"][request["requested_token_type"]]["kwargs"]
+        else:
+            fun = importer(self.config["token_exchange"]["policy"][""]["callable"])
+            kwargs = self.config["token_exchange"]["policy"][""]["kwargs"]
+
+        return fun(request, context=_context, kwargs=kwargs)
```
nsklikas, line 491: I think this should be wrapped in a try/except and return a `server_error` if any exception is raised

nsklikas, line 491: I think this should be:
```python
         return fun(request, context=_context, **kwargs)
```

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,249 @@ def post_parse_request(
...
+        token = _mngr.find_token(_session_info["session_id"], request["subject_token"])
+
+        resp = self.enforce_policy(request, token)
+        if isinstance(resp, TokenErrorResponse):
+            return resp
+
+        if token.is_active() is False:
+            return self.error_cls(
+                error="invalid_request", error_description="Subject token inactive"
+            )
```
nsklikas, line 434: First check if the token is active and then enforce the policy

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,249 @@ def post_parse_request(
...
+        except (KeyError, UnknownToken):
+            logger.error("Subject token invalid.")
+            return self.error_cls(
+                error="invalid_request",
+                error_description="Subject token invalid"
+            )
+
+        token = _mngr.find_token(_session_info["session_id"], request["subject_token"])
+
+        resp = self.enforce_policy(request, token)
```
nsklikas, line 427: We should in the client metadata for a policy as well and use that policy if found

src/oidcop/oidc/token.py
ctriant: `subject_token_types_supported` is a check for the integrity of the token exchange request, but shouldn't we check that the actual `token` is of the equivalent class? 

What about we add a check upon a allowed mapping like:
```python
subject_token_types_supported = {
            "urn:ietf:params:oauth:token-type:access_token": AccessToken,
            "urn:ietf:params:oauth:token-type:refresh_token": RefreshToken
        }

if (
    request["subject_token_type"] in subject_token_types_supported
    and (
        not isinstance(token,
                subject_token_types_supported[request["subject_token_type"]]
        )
    )
):
    return self.error_cls(
        error="invalid_request", error_description="Wrong token type"
    )
```

nsklikas: LGTM

src/oidcop/oidc/token.py
ctriant: I made this part of the default token exchange callable

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,253 @@ def post_parse_request(
...
+        if isinstance(resp, TokenErrorResponse):
+            return resp
+
+        if token.is_active() is False:
+            return self.error_cls(
+                error="invalid_request", error_description="Subject token inactive"
+            )
```
nsklikas, line 434: Check if the token is active before enforcing the policy

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,253 @@ def post_parse_request(
...
+
+        if (
+            "requested_token_type" in request
+            and request["requested_token_type"] in self.config["token_exchange"]["policy"]
+        ):
+            fun = importer(self.config["token_exchange"]["policy"][request["requested_token_type"]]["callable"])
+            kwargs = self.config["token_exchange"]["policy"][request["requested_token_type"]]["kwargs"]
+        else:
+            fun = importer(self.config["token_exchange"]["policy"][""]["callable"])
+            kwargs = self.config["token_exchange"]["policy"][""]["kwargs"]
```
nsklikas, line 469: What if that doesn't exist?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,253 @@ def post_parse_request(
...
+
+        return fun(request, context=_context, subject_token=token, kwargs=kwargs)
```
nsklikas, line 471: What happens if this raises an exception?

I think we should catch it and return a server error

nsklikas, line 471: Shouldn't this be 
```python
         return fun(request, context=_context, subject_token=token, **kwargs)
```
?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,253 @@ def post_parse_request(
...
+            )
+        return resp
+
+    def enforce_policy(self, request, token):
+        _context = self.endpoint.server_get("endpoint_context")
+
+        if not isinstance(token, (AccessToken, RefreshToken)):
+            return self.error_cls(
+                error="invalid_request", error_description="Wrong token type"
+            )
```
nsklikas, line 444: I think that this is no longer needed

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,254 @@ def post_parse_request(
...
+        ):
+            fun = importer(self.config["token_exchange"]["policy"][request["requested_token_type"]]["callable"])
+            kwargs = self.config["token_exchange"]["policy"][request["requested_token_type"]]["kwargs"]
+        else:
+            fun = importer(self.config["token_exchange"]["policy"][""]["callable"])
+            kwargs = self.config["token_exchange"]["policy"][""]["kwargs"]
+
+        return fun(request, context=_context, subject_token=token, kwargs=kwargs)
+
+    def token_exchange_response(self, access_token):
```
nsklikas, line 473: Let's not name the param `access_token`, the spec is outdated and we don't need to follow the same notation

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,274 @@ def post_parse_request(
...
+        else:
+            self.config = config
+
+    def post_parse_request(self, request, client_id="", **kwargs):
+        request = TokenExchangeRequest(**request.to_dict())
+
+        _context = self.endpoint.server_get("endpoint_context")
+        if "token_exchange" in self.endpoint.kwargs:
+            self.config["token_exchange"] = _context.cdb[request["client_id"]].get(
+                "token_exchange", self.endpoint.kwargs["token_exchange"])
```
nsklikas, line 383: Don't overwrite the object's config. 
```python
if "token_exchange" in _context.cdb[request["client_id"]]:
    self.config["token_exchange"] = _context.cdb[request["client_id"]]["token_exchange"]
else:
    config = self.endpoint.kwargs["token_exchange"]
```

Can `token_exchange` not exist in `self.endpoint.kwargs`? Would that allow us to enable token exchange only for specific clients while disabling it globally?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,274 @@ def post_parse_request(
...
+        if token.is_active() is False:
+            return self.error_cls(
+                error="invalid_request", error_description="Subject token inactive"
+            )
+
+        resp = self.enforce_policy(request, token)
+        if isinstance(resp, TokenErrorResponse):
+            return resp
+
+        return resp
```
nsklikas, line 431:
```python
return resp
```

No reason for this `if` to exist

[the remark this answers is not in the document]
ctriant: The `enforce_policy` method internally calls the callable provided in the configuration (by default `default_token_exchange_policy`) and returns the request after enforcing the policy (eg scope filtering).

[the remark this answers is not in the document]
nsklikas: We should document that behavior. At least let's describe the expected behavior of the callable in the documentation.

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,274 @@ def post_parse_request(
...
+        ):
+            callable = self.config["token_exchange"]["policy"][request["requested_token_type"]]["callable"]
+            kwargs = self.config["token_exchange"]["policy"][request["requested_token_type"]]["kwargs"]
+        else:
+            callable = self.config["token_exchange"]["policy"][""]["callable"]
+            kwargs = self.config["token_exchange"]["policy"][""]["kwargs"]
+
+        fn = importer(callable)
+        try:
+            return fn(request, context=_context, subject_token=token, **kwargs)
```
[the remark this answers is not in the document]
nsklikas: Maybe we should also pass the `session_info` since we have already retrieved it?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,281 @@ def post_parse_request(
...
+        ):
+            return self.error_cls(
+                error="invalid_request", error_description="Wrong token type"
+            )
+
+        if request["subject_token_type"] not in subject_token_types_supported:
+            return TokenErrorResponse(
+                error="invalid_request",
+                error_description="Unsupported subject token type",
+            )
```
nsklikas, line 472: Let's rename `self.total_subject_token_types_supported` to `self.subject_token_types_supported`

I think this approach is simpler:
```python
subject_token_types_supported = config.get(
    "subject_token_types_supported", self.subject_token_types_supported
)
subject_token_type = request["subject_token_type"]
if (
    subject_token_type not in subject_token_types_supported 
    or subject_token_types_supported[subject_token_type] != token.token_class
):
    return TokenErrorResponse(
        error="invalid_request",
        error_description="Unsupported subject token type",
    )
```

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,281 @@ def post_parse_request(
...
+                    "urn:ietf:params:oauth:token-type:access_token",
+                    "urn:ietf:params:oauth:token-type:refresh_token",
+                ],
+                "requested_token_types_supported": [
+                    "urn:ietf:params:oauth:token-type:access_token",
+                    "urn:ietf:params:oauth:token-type:refresh_token",
+                ],
+                "policy": {
+                    "": {
+                        "callable": "oidcop.oidc.token.default_token_exchange_policy",
```
nsklikas, line 384: Since the function is defined in this module let's use the actual function here, no reason to import it later.

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,281 @@ def post_parse_request(
...
+        try:
+            if (
+                "requested_token_type" in request
+                and request["requested_token_type"] in config["policy"]
+            ):
+                callable = config["policy"][request["requested_token_type"]]["callable"]
+                kwargs = config["policy"][request["requested_token_type"]]["kwargs"]
+            else:
+                callable = config["policy"][""]["callable"]
+                kwargs = config["policy"][""]["kwargs"]
```
nsklikas, line 502: This would be simpler I think:
```python
requested_token_type = request.get("requested_token_type", "")
if requested_token_type in config["policy"]:
    callable = config["policy"][requested_token_type]["callable"]
    kwargs = config["policy"][requested_token_type]["kwargs"]
else:
    return error
```

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,281 @@ def post_parse_request(
...
+        else:
+            _token_type = None
+
+        sid = _session_info["session_id"]
+        if request["client_id"] != _session_info["client_id"]:
+            authn_event = _session_info["grant"].authentication_event
+            _token_usage_rules = _context.authz.usage_rules(request["client_id"])
+            _exp_in = _token_usage_rules["refresh_token"].get("expires_in")
+            if _exp_in and "valid_until" in authn_event:
+                authn_event["valid_until"] = utc_time_sans_frac() + _exp_in
```
nsklikas, line 562: this seems wrong, why is this here?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,281 @@ def post_parse_request(
...
+
+            sid = _mngr.create_session(
+                authn_event=authn_event,
+                auth_req=request,
```
nsklikas, line 566: this isn't an authentication request, it's a token request, perhaps this should be `None`?

docs/source/contents/conf.rst
```diff
@@ -661,6 +663,120 @@ the following::
...
+==============
+Token exchange
+==============
+There are two possible ways to configure Token Exchange in OIDC-OP, globally and per-client.
+For the first case the configuration is passed in the Token Exchange handler throught the
+`urn:ietf:params:oauth:grant-type:token-exchange` dictionary in token's `grant_types_supported`.
+
+If present, the token exchange configuration must contain a `policy` object that describes a default
+policy `callable` and its `kwargs` through the `""` key. Different callables can be optionally
+defined for each token type supported.
```
nsklikas, line 675: Why MUST it contain the default `""` key?

src/oidcop/oidc/token.py
```diff
@@ -352,6 +363,281 @@ def post_parse_request(
...
+                error_description="Unauthorized scope requested",
+            )
+
+        try:
+            if (
+                "requested_token_type" in request
+                and request["requested_token_type"] in config["policy"]
+            ):
+                callable = config["policy"][request["requested_token_type"]]["callable"]
+                kwargs = config["policy"][request["requested_token_type"]]["kwargs"]
```
nsklikas, line 499: Use `subject_token_type` instead.

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +362,284 @@ def post_parse_request(
...
+            return TokenErrorResponse(
+                error="invalid_request",
+                error_description="Unsupported requested token type",
+            )
+
+        request_info = dict(scope=request.get("scope", []))
+        try:
+            check_unknown_scopes_policy(request_info, request["client_id"], _context)
+        except UnAuthorizedClientScope:
+            logger.error("Unauthorized scope requested.")
```
nsklikas, line 486: This should not be logged as an error

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +362,284 @@ def post_parse_request(
...
+                error="invalid_grant",
+                error_description="Unauthorized scope requested",
+            )
+
+        try:
+            subject_token_type = request.get("subject_token_type", "")
+            if subject_token_type not in config["policy"]:
+                subject_token_type = ""
+            callable = config["policy"][subject_token_type]["callable"]
+            kwargs = config["policy"][subject_token_type]["kwargs"]
```
nsklikas, line 497: Can we move this outside the `try` block and check if `subject_token_type` or `""` is in the `config["policy"]`, if not let's raise `oidcop.exception.ImproperlyConfigured`:
```python
subject_token_type = request["subject_token_type"]
if subject_token_type not in config["policy"]:
    if "" not in config["policy"]:
        raise ImproperlyConfigured(
            "subject_token_type {subject_token_type} missing from "
            "policy and no default is defined"
        )
    subject_token_type = ""
policy = config["policy"][subject_token_type]
...
```

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +362,284 @@ def post_parse_request(
...
+
+            if isinstance(callable, str):
+                fn = importer(callable)
```
nsklikas, line 500: wrap this in a separate `try/except` and raise `ImproperlyConfigured` with the appropriate message

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +362,284 @@ def post_parse_request(
...
+            else:
+                fn = callable
+            return fn(request, context=_context, subject_token=token, **kwargs)
+
+        except Exception:
```
nsklikas, line 505: log some error related to the exception for debuging

src/oidcop/session/manager.py
```diff
@@ -198,6 +200,41 @@ def create_grant(
...
+
+        :param scopes: Scopes
+        :param exchange_req:
+        :param user_id:
+        :param client_id:
+        :param sub_type:
+        :param token_usage_rules:
+        :return:
+        """
+        sector_identifier = exchange_request.get("sector_identifier_uri", "")
```
nsklikas, line 223: This parameter is not defined in the spec, do we need this?

src/oidcop/session/manager.py
```diff
@@ -198,6 +200,41 @@ def create_grant(
...
+
+        _claims = exchange_request.get("claims", {})
```
nsklikas, line 225: This parameter is not defined in the spec, do we need this?

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +362,284 @@ def post_parse_request(
...
+                error_description="Internal server error"
+            )
+
+    def token_exchange_response(self, token):
+        response_args = {}
+        response_args["access_token"] = token.value
+        response_args["scope"] = token.scope
+        response_args["issued_token_type"] = token.token_class
+        response_args["expires_in"] = token.usage_rules.get("expires_in", 0)
+        response_args["token_type"] = "bearer"
```
nsklikas, line 517: From https://datatracker.ietf.org/doc/html/rfc8693#section-2.2.1:

> token_type
   ...
   If the issued token is not an access token or 
   usable as an access token, then the "token_type" value "N_A" is
   used to indicate that an OAuth 2.0 "token_type" identifier is not
   applicable in that context.

src/oidcop/oauth2/token.py
nsklikas: Only this line (`fn = importer(callable)`) can raise no reason to have the rest in the `try/except` block

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+        kwargs = policy["kwargs"]
+
+        try:
+            if isinstance(callable, str):
+                fn = importer(callable)
+            else:
+                fn = callable
+        except Exception:
+            raise ImproperlyConfigured(
+                "Error importing {callable} policy callable"
```
nsklikas, line 503:
```python
f"Error importing {callable} policy callable"
```

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+            )
+
+        try:
+            return fn(request, context=_context, subject_token=token, **kwargs)
+        except Exception:
+            logger.error("Error while executing the {fn} policy callable")
```
nsklikas, line 509:
```python
        logger.error(f"Error while executing the {fn} policy callable")
```

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+                error_description="Subject token invalid"
+            )
+
+        token = _mngr.find_token(_session_info["session_id"], request["subject_token"])
+        _requested_token_type = request.get("requested_token_type",
+                                           "urn:ietf:params:oauth:token-type:access_token")
+
+        _token_class = _requested_token_type.split(":")[-1]
+        if _token_class == "access_token":
+            _token_type = _token_class
```
nsklikas, line 554: This must be `Bearer` or `DPoP`. We could use the same logic as in https://github.com/IdentityPython/oidc-op/pull/165/files#diff-701a8905a1282dde6d1b7325ab65768a840a3942e41112d24135688431564df3R255-R262 ?

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+            return self.error_cls(
+                error="invalid_request",
+                error_description="Subject token invalid"
+            )
+
+        token = _mngr.find_token(_session_info["session_id"], request["subject_token"])
+        _requested_token_type = request.get("requested_token_type",
+                                           "urn:ietf:params:oauth:token-type:access_token")
+
+        _token_class = _requested_token_type.split(":")[-1]
```
nsklikas, line 552: Can we make this dynamic? Could we use self.subject_types_supported_mapping (change it's name to `token_type_to_class` and make the `values` class instances?)

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+                user_id=_session_info["user_id"],
+                client_id=request["client_id"],
+                token_usage_rules=_token_usage_rules,
+            )
+
+            try:
+                _session_info = _mngr.get_session_info(
+                    session_id=sid, grant=True)
+            except Exception:
+                logger.error("Error retrieving token exchabge session information")
```
nsklikas, line 574: `retrieving token exchaNge`

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+        try:
+            new_token = self._mint_token(
+                token_class=_token_class,
+                grant=_session_info["grant"],
+                session_id=sid,
+                client_id=request["client_id"],
+                based_on=token,
+                scope=request.get("scope"),
+                token_args={
+                    "resources":request.get("resource"),
```
nsklikas, line 589: Shouldn't `audience` be in the `token_args` as well?

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+                error_description="Token Exchange not allowed with that token",
+                )
+
+        return self.token_exchange_response(token=new_token)
+
+def default_token_exchange_policy(request, context, subject_token, **kwargs):
+    if "resource" in request:
+        resource = kwargs.get("resource", [])
+        if not resource:
+            pass
```
nsklikas, line 606: This is unnecessary

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+                error="invalid_target", error_description="Unknown resource"
+            )
+
+    if "audience" in request:
+        if request["subject_token_type"] == "urn:ietf:params:oauth:token-type:refresh_token":
+            return TokenErrorResponse(
+                error="invalid_target", error_description="Refresh token has single owner"
+        )
+        audience = kwargs.get("audience", [])
+        if not audience:
```
nsklikas, line 618: This is unnecessary

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+        "requested_token_type" in request
+        and request["requested_token_type"] == "urn:ietf:params:oauth:token-type:refresh_token"
+    ):
+        if "offline_access" not in subject_token.scope:
+            return TokenErrorResponse(
+                error="invalid_request",
+                error_description=f"Exchange {request['subject_token_type']} to refresh token forbbiden",
+            )
+
+    scopes = list(set(request.get("scope", ["openid"])).intersection(kwargs.get("scope", ["openid"])))
```
nsklikas, line 640: Why set the `scope` default value to `openid`?

src/oidcop/oauth2/token.py
```diff
@@ -358,6 +363,289 @@ def post_parse_request(
...
+            )
+
+    if "audience" in request:
+        if request["subject_token_type"] == "urn:ietf:params:oauth:token-type:refresh_token":
+            return TokenErrorResponse(
+                error="invalid_target", error_description="Refresh token has single owner"
+        )
+        audience = kwargs.get("audience", [])
+        if not audience:
+            pass
```
nsklikas, line 619: This `if` clause is unnecessary

src/oidcop/session/grant.py
```diff
@@ -444,22 +444,22 @@ class ExchangeGrant(Grant):
...
         usage_rules: Optional[dict] = None,
+        exchange_request: str = "",
+        original_session_id: str = "",
         issued_at: int = 0,
         expires_in: int = 0,
         expires_at: int = 0,
         revoked: bool = False,
         token_map: Optional[dict] = None,
-        users: list = None,
+        users: list = None
```
nsklikas, line 458: What is the use of this parameter? Maybe we should remove it?

src/oidcop/oauth2/token.py
```diff
@@ -359,6 +367,270 @@ def post_parse_request(
...
+        request_info = dict(scope=request.get("scope", []))
+        try:
+            check_unknown_scopes_policy(request_info, request["client_id"], _context)
+        except UnAuthorizedClientScope:
+            return self.error_cls(
+                error="invalid_grant",
+                error_description="Unauthorized scope requested",
+            )
+
+        subject_token_type = request["subject_token_type"]
```
nsklikas, line 473: This is already defined in https://github.com/IdentityPython/oidc-op/pull/165/files#diff-701a8905a1282dde6d1b7325ab65768a840a3942e41112d24135688431564df3R443

src/oidcop/oauth2/token.py
```diff
@@ -359,6 +367,270 @@ def post_parse_request(
...
+        except Exception as e:
+            logger.error(f"Error while executing the {fn} policy callable: {e}")
+            return self.error_cls(error="server_error", error_description="Internal server error")
+
+    def token_exchange_response(self, token):
+        response_args = {}
+        response_args["access_token"] = token.value
+        response_args["scope"] = token.scope
+        response_args["issued_token_type"] = token.token_class
+        response_args["expires_in"] = token.usage_rules.get("expires_in", 0)
```
nsklikas, line 505: Let's use the same logic as in https://github.com/IdentityPython/oidc-op/blob/master/src/oidcop/oauth2/token.py#L166:
```python
if token.expires_at:
    _response["expires_in"] = token.expires_at - utc_time_sans_frac()
```

src/oidcop/oauth2/token.py
```diff
@@ -359,6 +367,270 @@ def post_parse_request(
...
+                error_description="Token Exchange not allowed with that token",
+            )
+
+        return self.token_exchange_response(token=new_token)
+
+
+def default_token_exchange_policy(request, context, subject_token, **kwargs):
+    if "resource" in request:
+        resource = kwargs.get("resource", [])
+        if not len(set(request["resource"]).intersection(set(resource))):
```
nsklikas, line 595: This is cleaner:
```python
if set(request["resource"]).issubset(set(resource))
```

src/oidcop/oauth2/token.py
```diff
@@ -359,6 +367,270 @@ def post_parse_request(
...
+            return TokenErrorResponse(error="invalid_target", error_description="Unknown resource")
+
+    if "audience" in request:
+        if request["subject_token_type"] == "urn:ietf:params:oauth:token-type:refresh_token":
+            return TokenErrorResponse(
+                error="invalid_target", error_description="Refresh token has single owner"
+            )
+        audience = kwargs.get("audience", [])
+        if audience and not len(set(request["audience"]).intersection(set(audience))):
```
nsklikas, line 604: same (use `issubset`)

src/oidcop/session/grant.py
```diff
@@ -471,7 +470,8 @@ def __init__(
             token_map=token_map,
         )
 
-        self.users = users or []
         self.usage_rules = {
             "access_token": {"supports_minting": ["access_token"], "expires_in": 60}
         }
```
nsklikas, line 475: I don't think this is still needed.

src/oidcop/session/manager.py
```diff
@@ -198,6 +201,33 @@ def create_grant(
...
+        :param scopes: Scopes
+        :param exchange_req:
+        :param user_id:
+        :param client_id:
+        :param sub_type:
+        :return:
+        """
+
+        grant = ExchangeGrant(
+            scope=scopes, original_session_id=original_session_id, exchange_request=exchange_request
```
nsklikas, line 225: Some parameters are missing here, i.e. `sub`, `usage_rules` (see https://github.com/IdentityPython/oidc-op/blob/master/src/oidcop/session/manager.py#L186)

nsklikas:
Run darker on this. Other than this LGTM. 

@rohe @peppelinux what do you think?

src/oidcop/oauth2/token.py
```diff
@@ -359,6 +367,270 @@ def post_parse_request(
...
+            return fn(request, context=_context, subject_token=token, **kwargs)
+        except Exception as e:
+            logger.error(f"Error while executing the {fn} policy callable: {e}")
+            return self.error_cls(error="server_error", error_description="Internal server error")
+
+    def token_exchange_response(self, token):
+        response_args = {}
+        response_args["access_token"] = token.value
+        response_args["scope"] = token.scope
+        response_args["issued_token_type"] = token.token_class
```
nsklikas, line 503: This must be the `issued_token_type` that was requested not the one on the class (`urn:ietf:params:oauth:token-type:access_token` vs `access_token`)

src/oidcop/oidc/token.py
rohe: So far we've uses a `verify` method to check if a message is correct. I'd rather see what we continue doing that.
Hence the `verify` method in TokenExchangeRequest should be extended to include these tests.

rohe:
Nice set of tests !

nsklikas:
We merged this in our private fork, here are some things that need to be done:
- Check that the new scopes are a subset of the allowed scopes for the client
- Set the expires_at of the new token to be the same as the original token's. This would stop a client from getting an access token from another client and using token exchange to refresh it indefinitely.

peppelinux:
@nsklikas we just have two conflicting files right now :)

melanger:
@peppelinux @nsklikas Hello, are you planning to merge this in public repository? I would really like to have token exchange in SATOSA.

ctriant:
> @peppelinux @nsklikas Hello, are you planning to merge this in public repository? I would really like to have token exchange in SATOSA.

Hello! It is already supported in idpy-oidc. https://github.com/IdentityPython/idpy-oidc/tree/main

rohe:
It's supported in idpy-oidc but the official SATOSA version at IdentityPython is still not based on idpy-oidc.

melanger:
The default OIDC frontend in SATOSA uses pyop and oic, but I am using satosa-oidcop which uses this library (oidc-op).

> It's supported in idpy-oidc but the official SATOSA version at IdentityPython is still not based on idpy-oidc.

So which library is the "new" one? I though it is this one.

rohe:
idpy-oidc is where the action is.

melanger:
> idpy-oidc is where the action is.

So let me recap:

* SATOSA contains [OpenIDConnectFrontend](https://github.com/IdentityPython/SATOSA/blob/master/src/satosa/frontends/openid_connect.py) based on [pyop](https://github.com/IdentityPython/pyop), that's an "old" one
* [satosa-oidcop](https://github.com/UniversitaDellaCalabria/SATOSA-oidcop) provides [OidcOpFrontend](https://github.com/UniversitaDellaCalabria/SATOSA-oidcop/blob/main/satosa_oidcop/idpy_oidcop.py) based on [oidc-op](https://github.com/IdentityPython/oidc-op) (this library), which is a "less old" one
* the current library is [idpy-oidc](https://github.com/IdentityPython/idpy-oidc), but there is no SATOSA frontend which uses the library

What are the development plans? Are you going to support oidc-op until idpy-oidc is mature enough and integrated with SATOSA?

rohe:
Good questions all.
I'm responsible for the OIDC/OAuth2 libraries within IdentityPython and I would like SATOSA to switch to using idpy-oidc as soon as possible.
I think there is a SATOSA frontend, based on idpy-oidc, in the EduTeams implementation but it isn't public yet.

PR #166, opened 2021-12-15
Use configuration classes from oidcmsg.

rohe:
Need to have oidcmsg 1.5.4 .
utc_time_sans_frac instead of time.time() since the later is unspecified when it comes to timezone.
Code dealing with configuration parsing that was present both in oidc-op and oidcmsg is now removed from oidc-op.


peppelinux (approved):
Please add in the PR description also the Configuration clean up, we need it in the changelog of the release

thx ;)

rohe:
Done

nsklikas (approved):

PR #167, opened 2021-12-15
Project version behind pypi version. This to put them in sync.

rohe:
For some reason library and pypi version was out of sync.


rohe:
May need to add more.

PR #173, opened 2022-01-26
Support both list and dict

nsklikas:
The default type for the "always" claims is `dict`, this seems to have broken in some recent change.


peppelinux (approved):
sorry for the late, looks good

rohe (approved):

PR #175, opened 2022-01-29
An add_on that allows adding extra arguments to a response.

rohe:
There to support https://datatracker.ietf.org/doc/draft-ietf-oauth-iss-auth-resp .


peppelinux:
That's great @rohe 

I believe that we really need a point in readme related to rfc supported and few words in the doc

I can do that but I'll take more than two week

If you Need to merge this asap, go ahead but open and issue to this PR, we'll cover this aspects onanother milestone

peppelinux (approved):

PR #176, opened 2022-01-31
Response might not be a Message instance.

rohe:
What's returned as bases for a response might not be a Message instance.


nsklikas (approved):

peppelinux:
Sorry for the late, I was sure to have already approved this, thank you

PR #182, opened 2022-02-01
A number of bud fixes and new functionality

rohe:


src/oidcop/__init__.py
```diff
@@ -1,6 +1,6 @@
 import secrets
 
-__version__ = "2.3.4"
+__version__ = "2.3.5"
```
peppelinux, line 3: Considering that we have an additional feature like iss identification, would we move to 2.4.0 instead or 2.3.5 if good right now?

rohe: OK

PR #183, opened 2022-02-03
Add more clear client/token authn exceptions

nsklikas:
- Add more/clearer exceptions so that it is easier for the caller to handle client authentication failure. 
- Raise `BearerTokenAuthenticationError` instead of `ValueError`.
- `get_client_id_from_token` defaults to `None`.

### DIFF
--- a/example/flask_op/views.py
+++ b/example/flask_op/views.py
@@ -18,8 +18,8 @@ from oidcmsg.oidc import AuthorizationRequest
 import werkzeug
 
 from oidcop.exception import FailedAuthentication
-from oidcop.exception import InvalidClient
-from oidcop.exception import UnknownClient
+from oidcop.exception import ClientAuthenticationError
+from oidcop.exception import TokenAuthenticationError
 from oidcop.oidc.token import Token
 
 # logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


src/oidcop/endpoint.py
```diff
@@ -206,14 +206,6 @@ def parse_request(
             request=req, client_id=_client_id, http_info=http_info, **kwargs
         )
 
-    def get_client_id_from_token(
```
peppelinux, line 209: this is a callable and not a property, at line 94 you create instead an attribute

are we sure that's won't be faulty?

example/flask_op/views.py
```diff
@@ -224,12 +224,18 @@ def service_endpoint(endpoint):
     if request.method == 'GET':
         try:
             req_args = endpoint.parse_request(request.args.to_dict(), http_info=http_info)
-        except (InvalidClient, UnknownClient) as err:
```
peppelinux, line 227: Are we sure that invalid and unknow client won't reiase here?

peppelinux (commented):
I think it's worth of it!

rohe (approved):

src/oidcop/endpoint.py
nsklikas: I changed it so that it's only defined in the classes that need it

example/flask_op/views.py
nsklikas: That is the point, `InvalidClient` and `UnknownClient` are subclasses of `ClientAuthenticationError`. They will be handled by the new `except` statement.

peppelinux (approved):

PR #186, opened 2022-02-09
Separate grant_types_supported and grant_types_helpers

nsklikas:
Separate `grant_types_supported` and `grant_types_helpers`.

The reason behind this is that we may want to initialize some grant_type helper, but no support that grant type globally. For example, we may want to support token_exchange for only one client, the helper must be initialized, but the grant type must not be supported globally.


rohe (approved):

peppelinux (approved):
Ok, good!

@nsklikas please use a conventional commit with a BREAKING CHANGE label to show to the user that the release that will bring this change will introduce problems with the preexisting configurations

PR #188, opened 2022-02-16
Catch exception, fix error message

nsklikas:
Catch unknown token if raised and fix the error response message.

### DIFF
--- a/src/oidcop/oauth2/token.py
+++ b/src/oidcop/oauth2/token.py
@@ -332,9 +332,9 @@ class RefreshTokenHelper(TokenEndpointHelper):
         _mngr = _context.session_manager
         try:
             _session_info = _mngr.get_session_info_by_token(request["refresh_token"], grant=True)
-        except KeyError:
-            logger.error("Access Code invalid")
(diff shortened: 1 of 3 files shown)


src/oidcop/oauth2/token.py
```diff
@@ -332,9 +332,9 @@ def post_parse_request(
         _mngr = _context.session_manager
         try:
             _session_info = _mngr.get_session_info_by_token(request["refresh_token"], grant=True)
-        except KeyError:
-            logger.error("Access Code invalid")
```
peppelinux, line 336: I trust you as usual @nsklikas, just a question regarding why you dropped Access Code error for a refresh one.

Probably the old one was wrong and copletely out of scope there? Or do we're moving on a breaking change?

nsklikas: The old one was wrong as this is the RefreshTokenHelper, so the token we are expecting is a refresh token. This was probably copied from the AccessTokenHelper and slipped past review

rohe (approved):

PR #189, opened 2022-02-19
Erasing tokens

rohe:
If configured so the software will store information about a token outside the running code before removing it from the Grant is was created within.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-oidcmsg>=1.4.0
+oidcmsg>=1.6.0
 pyyaml
 jinja2>=2.11.3
 responses>=0.13.0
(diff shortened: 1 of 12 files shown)


peppelinux:
Ok, before doing a review we need to have a new release of oidcmsg and oidcrp if you agree

I made my part here:
https://github.com/IdentityPython/JWTConnect-Python-OidcMsg/pull/56

rohe:
Yes, we do need a new release of oidcmsg.

peppelinux (approved):

peppelinux:
I'd consider to have first the oidcmsg release and then oidcrp and oidcop

PRs approved, if help is needed just tell me

rohe:
For sure, oidcmsg first, then oidcop&oidcrp.

PR #196, opened 2022-03-03
Fix utils wrong authorizationrequest import

### LINKED ISSUE #195: Fastapi example is broken
The Fastapi example fails to start due to several issues:

- Broken  AuthorizationRequest import in utils.py
- App ends prematurely without logging error due to initialization code in `startup` event handler
- Broken config

NubiaNetworks:
Fixes #195

### DIFF
--- a/example/fastapi/config.json
+++ b/example/fastapi/config.json
@@ -1,330 +1,370 @@
 {
-  "add_on": {
-    "pkce": {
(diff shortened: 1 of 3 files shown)


peppelinux (approved):

rohe (approved):
I'm not a FastAPI user so I'm not the person to have any view on that side.
I'll have to trust Giuseppe for checking that part.

PR #198, opened 2022-03-09
Remove client_authn_method from provider

nsklikas:
Remove the `client_authn_method` from the provider info endpoint. It is not described in the spec and it is pretty buggy, as each endpoint can over-write the methods of the other endpoints, which means that the result relies on the iteration order.

If we want to expose the client authentication methods supported for the other endpoints,  we should do it per endpoint, e.g. `introspection_endpoint_auth_methods_supported` (like `token_endpoint_auth_methods_supported`, which is the only one described in the discovery specification)

### DIFF
--- a/src/oidcop/server.py
+++ b/src/oidcop/server.py
@@ -37,7 +37,8 @@ def do_endpoints(conf, server_get):
 
 
 def get_capabilities(conf, endpoints):
-    _cap = conf.get("capabilities", {})
+    ignored_capabilities = ["client_authn_method"]
+    _cap = dict(conf.get("capabilities", {}))
     if _cap is None:
         _cap = {}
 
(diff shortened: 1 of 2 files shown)


rohe:
I think we should spend some time on making get_capabilities better, rather than this fix.

nsklikas:
Perhaps each endpoint should have a method that will generate it's capabilities and then the server will fetch/merge all of them?

What happens right now is similar to that but the endpoint capabilities are overloaded, as they are used to both configure the endpoints behavior, with `client_authn_method`, and to create the endpoint's capabilities that are exposed in the discovery endpoint. IMO the user should have to configure the endpoint and then the endpoints capabilities should be generated from the provided configuration, which then the server will merge to create the output of the discovery endpoint. We could also add an extra field to the endpoints for overriding the discovery endpoint's capabilities, in case the user wants some custom behavior.

rohe:
My thoughts are that the endpoints should register their capabilities with the server. Probably best implemented with the callback method you talked about.
Which provider configuration parameters that belongs to which endpoint is defined in the code. These parameters may have default values which depends on the capabilities of the software package but are ultimately set by configuration.
Setting the provider configuration parameters should probably best be kept apart from other endpoint capabilities (like client_authn_methods).

